PR #106, opened 2021-04-21
usage fix

GeekSheikh:
There's a bug in scalaj for user-agent data in headers. Applied the work around and validated function. Working.

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/ApiCall.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/ApiCall.scala
@@ -26,6 +26,12 @@ class ApiCall(env: ApiEnv) extends SparkSessionWrapper {
   private var _status: String = "SUCCESS"
   private var _errorFlag: Boolean = false
   private val mapper = JsonUtils.defaultObjectMapper
+  private val httpHeaders = Seq[(String, String)](
+    ("Content-Type", "application/json"),
+    ("Charset",  "UTF-8"),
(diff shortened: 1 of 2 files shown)


src/main/scala/com/databricks/labs/overwatch/ApiCall.scala
```diff
@@ -132,12 +138,7 @@ class ApiCall(env: ApiEnv) extends SparkSessionWrapper {
     logger.log(Level.INFO, s"Loading $req -> query: $getQueryString")
     try {
       val result = Http(req + getQueryString)
-        .headers(Map[String, String](
-          "Content-Type" -> "application/json",
-          "Charset"-> "UTF-8",
-          "User-Agent" -> s"databricks-labs-overwatch/${env.packageVersion}",
-          "Authorization" -> s"Bearer ${env.rawToken}"
-        ))
+        .copy(headers = httpHeaders)
```
alexott, line 141: why not construct `HttpRequest` directly instead of creating it indirectly & then copying? https://github.com/scalaj/scalaj-http/blob/master/src/main/scala/scalaj/http/Http.scala#L791

GeekSheikh: It's a bug in scalaJ. I tested several different options and this was the one that worked. See the bug below.
https://github.com/scalaj/scalaj-http/issues/113

alexott (approved):
lgtm

PR #107, opened 2021-04-21
dropping docs from master lineage

GeekSheikh:
Docs are served to the gh-pages from the public-docs branch thus it's confusing and duplicative to have them in two places since public docs will not be maintained in the master lineage. Still some work to perfect the docs process but this will kick the can a little further down the road.


alexott (approved):
lgtm

PR #110, opened 2021-04-26
external tables enabled

GeekSheikh:
Utilizing managed presents a risk if a user (even from a different workspace) calls `drop database overwatch cascade` in that all data within the database is deleted. Enabling external tables and presenting a WARNING if users specify a data target within /user/hive/warehouse prefix.


alexott (commented):
It could be simpler to leave `saveAsTable` but add the option to specify the path outside:

From Spark [docs](https://spark.apache.org/docs/latest/sql-data-sources-load-save-functions.html#saving-to-persistent-tables)
> For file-based data source, e.g. text, parquet, json, etc. you can specify a custom table path via the path option, e.g. `df.write.option("path", "/some/path").saveAsTable("t")`. W**hen the table is dropped, the custom table path will not be removed and the table data is still there**. If no custom table path is specified, Spark will write data to a default table path under the warehouse directory. When the table is dropped, the default table path will be removed too.

alexott (approved):
lgtm

PR #114, opened 2021-04-29
Bugfix minor various

GeekSheikh:
There were minor errors including

- browserHostName tag not available from notebookContext when running as job with main class
- empty audit log table resulted in a hard fail not soft error
- target paths were case sensitive instead of matching hive default behavior of lowercase only.

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/pipeline/Initializer.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/pipeline/Initializer.scala
@@ -417,7 +417,7 @@ object Initializer extends SparkSessionWrapper {
     logger.log(Level.INFO, "Initializing Config")
     val config = new Config()
     val orgId = if (dbutils.notebook.getContext.tags("orgId") == "0") {
-      dbutils.notebook.getContext.tags("browserHostName").split("\\.")(0)
+      dbutils.notebook.getContext.apiUrl.get.split("\\.")(0).split("/").last
     } else dbutils.notebook.getContext.tags("orgId")
     config.setOrganizationId(orgId)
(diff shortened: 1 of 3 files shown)


src/main/scala/com/databricks/labs/overwatch/pipeline/Initializer.scala
```diff
@@ -417,7 +417,7 @@ object Initializer extends SparkSessionWrapper {
     logger.log(Level.INFO, "Initializing Config")
     val config = new Config()
     val orgId = if (dbutils.notebook.getContext.tags("orgId") == "0") {
-      dbutils.notebook.getContext.tags("browserHostName").split("\\.")(0)
+      dbutils.notebook.getContext.apiUrl.get.split("\\.")(0).split("/").last
```
alexott, line 420: we can get orgId & Cloud from the Spark tags

alexott (approved):

src/main/scala/com/databricks/labs/overwatch/pipeline/Initializer.scala
GeekSheikh: Right, I'm just not entirely sure which of them get populated from within the different job types (i.e. notebook vs spark submit vs call to Main class). I will do some testing. Opening [issue](https://github.com/databrickslabs/overwatch/issues/115) to track this and will likely need to do even more due diligence for E2.

`spark.databricks.clusterUsageTags.clusterOwnerOrgId`
`spark.databricks.clusterUsageTags.cloudProvider`
`spark.databricks.clusterUsageTags.clusterAllTags`

PR #125, opened 2021-05-20
schema hotfix for jobs_snapshot_bronze

GeekSheikh:
hot fix to remove schema complexity in jobs_snapshot_bronze.

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala
@@ -554,6 +554,11 @@ trait BronzeTransforms extends SparkSessionWrapper {
             .otherwise(col("Stage ID"))
         } else col("Stage ID")
 
+        val stageAttemptIDColumnOverride: Column = if (baseEventsDF.columns.contains("stageAttemptId")) {
+          when('Event === "org.apache.spark.scheduler.SparkListenerSpeculativeTaskSubmitted", col("stageAttemptId"))
+            .otherwise(col("Stage Attempt ID"))
+        } else col("Stage Attempt ID")
+
         val rawScrubbed = if (baseEventsDF.columns.count(_.toLowerCase().replace(" ", "") == "stageid") > 1) {
           SchemaTools.scrubSchema(baseEventsDF
             .withColumn("progress", progressCol)
@@ -562,7 +567,8 @@ trait BronzeTransforms extends SparkSessionWrapper {
             .withColumn("SparkContextId", split('filename, "/")('pathSize - lit(2)))
             .withColumn("clusterId", split('filename, "/")('pathSize - lit(5)))
             .withColumn("StageID", stageIDColumnOverride)
-            .drop("pathSize", "Stage ID", "stageId")
+            .withColumn("StageAttemptID", stageAttemptIDColumnOverride)
+            .drop("pathSize", "Stage ID", "stageId", "Stage Attempt ID", "stageAttemptId")
             .withColumn("filenameGroup", groupFilename('filename))
           )
         } else {
--- a/src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala
@@ -439,7 +439,7 @@ trait BronzeTransforms extends SparkSessionWrapper {
     val byCluster = array_join(slice(segmentArray, 1, 3), "/").alias("byCluster")
     val byClusterHost = array_join(slice(segmentArray, 1, 5), "/").alias("byDriverHost")
     val bySparkContextID = array_join(slice(segmentArray, 1, 6), "/").alias("bySparkContext")
-    struct(filename, byCluster, byClusterHost, bySparkContextID).alias("filnameGroup")
+    struct(filename, byCluster, byClusterHost, bySparkContextID)
   }
 
   def generateEventLogsDF(database: Database,
--- a/src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala
@@ -146,18 +146,29 @@ trait BronzeTransforms extends SparkSessionWrapper {
 
   }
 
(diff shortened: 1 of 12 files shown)


src/main/scala/com/databricks/labs/overwatch/pipeline/Pipeline.scala
```diff
@@ -142,7 +142,7 @@ class Pipeline(_workspace: Workspace, _database: Database,
       spark.catalog.tableExists(config.databaseName, "pipeline_report")) {
       val w = Window.partitionBy('organization_id, 'moduleID).orderBy('Pipeline_SnapTS.desc)
       spark.table(s"${config.databaseName}.pipeline_report")
-        .filter('Status === "SUCCESS" || 'Status.startsWith("EMPTY"))
+        .filter('status === "SUCCESS" || 'status.startsWith("EMPTY"))
```
GeekSheikh, line 145: lower case "s" required. If failure occurs in spark_events_bronze where the spark SQL column case sensitivity is enabled, this could error.

src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala
```diff
@@ -146,18 +146,29 @@ trait BronzeTransforms extends SparkSessionWrapper {
 
   }
 
+  protected def cleanseRawJobsSnapDF(cloudProvider: String)(df: DataFrame): DataFrame = {
```
GeekSheikh, line 149: primary fix. Utilize similar logic to cleanse the raw jobs snapshot nasty fields.

src/main/scala/com/databricks/labs/overwatch/utils/Upgrade.scala
```diff
@@ -15,8 +15,53 @@ import scala.concurrent.forkjoin.ForkJoinPool
 object Upgrade extends SparkSessionWrapper {
 
   private val logger: Logger = Logger.getLogger(this.getClass)
+  import spark.implicits._
 
-  case class UpgradeStatus(db: String, tbl: String, actionTake: Option[String], errorMsg: Option[String])
+  def upgrade0412(prodWorkspace: Workspace): Dataset[UpgradeReport] = {
```
GeekSheikh, line 20: There is a schema change so a schema upgrade will be required if jobs_scope was run prior to 0.4.12.

src/main/scala/com/databricks/labs/overwatch/utils/Upgrade.scala
```diff
@@ -15,8 +15,53 @@ import scala.concurrent.forkjoin.ForkJoinPool
...
+    try {
+    if (!jobsSnapshotTarget.exists) {
+      val errMsg = s"The jobs_snapshot_bronze source does not exist in the workspace's configured database: " +
+        s"${bronzePipeline.config.databaseName}. This upgrade is only necessary if this source was created by " +
+        s"a version < 0.4.2"
+      logger.log(Level.ERROR, errMsg)
+      throw new UpgradeException(errMsg, jobsSnapshotTarget)
+    } else {
+      val jobSnapshotDF = jobsSnapshotTarget.asDF(withGlobalFilters = false)
+      val cloudProvider = bronzePipeline.config.cloudProvider
```
uucico, line 33: Is there any chance of getting this as an empty string? I know of some safeguards in validateAuditLogConfigs, and registerWorkspaceMeta (for apienv), but maybe when upgrading an assert somewhere to be defensive?

src/main/scala/com/databricks/labs/overwatch/pipeline/Initializer.scala
```diff
@@ -414,18 +426,24 @@ object Initializer extends SparkSessionWrapper {
...
+    logger.log(Level.INFO, "Initializing Workspace")
+    val workspace = Workspace(database, config)
+
+
+    workspace
+  }
+
+  private[overwatch] def apply(args: Array[String], debugFlag: Boolean, isSnap: Boolean): Workspace = {
+
+    val config = initConfigState(debugFlag)
```
GeekSheikh, line 446: all changes in this file were to enable dbConnect. They were pulled from the Kitana pipeline work in 0.4.2

src/main/scala/com/databricks/labs/overwatch/pipeline/Pipeline.scala
```diff
@@ -138,11 +138,20 @@ class Pipeline(_workspace: Workspace, _database: Database,
...
     logger.log(Level.INFO, "INIT: Pipeline Run")
+    val dbMeta = spark.sessionState.catalog.getDatabaseMetadata(config.databaseName)
+    val dbProperties = dbMeta.properties
+    val overwatchSchemaVersion = dbProperties.getOrElse("SCHEMA", "BAD_SCHEMA")
+    if (overwatchSchemaVersion != config.overwatchSchemaVersion) {
+      throw new BadConfigException(s"The overwatch DB Schema version is: $overwatchSchemaVersion but this" +
+        s" version of Overwatch requires ${config.overwatchSchemaVersion}. Upgrade Overwatch Schema to proceed " +
+        s"or drop existing database and allow Overwatch to recreate.")
+    }
+
```
GeekSheikh, line 149: moved this from Initializer.scala. I want to be able to instantiate the workspace with a different schema but not run a pipeline.

src/main/scala/com/databricks/labs/overwatch/utils/Upgrade.scala
GeekSheikh: Are you referring to the cloudProvider being an empty string? Under the current design I see no way it could be null and if it is there will be bigger problems elsewhere.

uucico: Yep. I only saw a check of it in validateAuditLogConfigs (it's in Initializer.scala), no other assertions or checks in the initialization, so I was thinking if we could get a wrong config (just in the upgrade process) for some reason, this would corrupt the field name.

GeekSheikh: The upgrade process still goes through the Initializer. The Initializer is required to get a workspace, thus the check would still be there to validate that this field is not null.

uucico (approved):
LGTM

chase-edwards-db:
This all looks good to me.

PR #149, opened 2021-06-30
Explicitly normalize non-DBFS URIs

alexott:
There was an error by customer where the URI was built by string concatenation, and
resulted in the string with `//` inside, and that lead to the errors on the second ingest
because URI was normalized by metastore, and comparison of original URI with URI from
metastore failed


GeekSheikh:
sorry I didn't see this before I had compiled the release jar. will approve after 042 tag created.

GeekSheikh (approved):
lgtm

Issue #134, opened 2021-06-07
[EPIC] - Publish DBSQL Gold Entities


GeekSheikh:
Originally, SQL Analytics Endpoints were identified as specialized clusters but they have since been modified into an entirely separate entity. As such, at this time, no SQL analytics data is being captured in the clusters_snapshot table but further investigation is necessary to determine how the audit logs have changed.

Targeting this resolution for next release 0.4.2

## Scope and Steps

1.  Get the dataframe from API . This will be defined in Workspace.scala. It will be similar to that of clusers_snapshot_bronze and jobs_snapshot_bronze this will act as a lookup for missing audit data in Silver
2. Define a endPointsTarget.  Just make sure it's defined as a Bronze target
3. Create a module inside Bronze.   Make sure it has an appropriate module ID and name. The schema scrubber implementation should NOT be INSIDE BRONZE -- but rather a function defined inside the trait, BronzeTransforms -- as needed
4. You create a process that implements ETL Definition inside Bronze.scala . You create a process that implements ETL Definition inside Bronze.scala -- this is what will actually call the functions in the BronzeTransformations that perform the schema cleansing, etc
5. Create a pipeline to execute the module and load the Target.  Bronze() is a pipeline so all you need to do is check the Overwatch scope to see if sqlAnalytics is present -- if it is, execute the process defined in Step 4, otherwise skip
which leads to step 0-ish which is likely add sqlAnalytics scope to the Overwatch Scope enumeration object and validate any requisite dependencies (there may not be any in bronze)

GeekSheikh:
Below is the code references to the issue and this will follow through to costs by cluster type in the gold layer.

[isSQLAnalytics Boolean]( https://github.com/databrickslabs/overwatch/blob/993ea86e956e7e01453fc58de33e2543392ea957/src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala#L443)

GeekSheikh:
Query History was added in 0.7.0 endpoints and other dbsql dims will be coming soon.

PR #150, opened 2021-07-01
Fix for SQL Endpoints API Change (issue #134)

jiteshsoni-db:
## EndPoint Documentation 
https://databrickslabs.github.io/overwatch/

#### What are we trying to solve
https://github.com/databrickslabs/overwatch/issues/134

### What needs to be captured 

1. states of endpoints through time
2. usage of endpoints through time
3. spec of endpoints through time
4. queries history, etc

once you have the inventory, create example DFs of what everything is going to look like in bronze (raw data) -- then see how it moves to silver and gold such that a user can see

1. the queries
2. the endpoints specs through time and how they've changed
3. the endpoint states through time
4. the costs
5. the query runtime metrics

yeah use the APIs and the Audit logs to essentially data mine out all the data points you have and create your inventory of what you have / don't have...from there we can work on the gold entities and start transforming the data to get there. I usually do most of this in notebooks to start, then I copy and paste the ETL code into Overwatch and pass in vars

## Ingestion points for Databricks SQL
We will need to consume data from multiple API endpoints and Audit logs.

List of End points to consume from:

1. "sql/endpoints"
2. "sql/history/queries"  The query string for the get request is include_metrics=true.

 

Audit logs:

1. Query audit_log_bronze table with alteast three different type of queries becuase Databricks SQL has been renamed. Example env https://adb-2222170229861029.9.azuredatabricks.net/?o=2222170229861029#setting/clusters/0608-082040-snore201/configuration

Use the database name: overwatch_etl

## Tasks

- [x]  1.  Get the dataframe from API . This will be defined in Workspace.scala. It will be similar to that of clusers_snapshot_bronze and jobs_snapshot_bronze this will act as a lookup for missing audit data in Silver
- [x] 2. Define a endPointsTarget.  Just make sure it's defined as a Bronze target
- [ ] 3. Create a module inside Bronze.   Make sure it has an appropriate module ID and name. The schema scrubber implementation should NOT be INSIDE BRONZE -- but rather a function defined inside the trait, BronzeTransforms -- as needed
- [ ] 4. You create a process that implements ETL Definition inside Bronze.scala . You create a process that implements ETL Definition inside Bronze.scala -- this is what will actually call the functions in the BronzeTransformations that perform the schema cleansing, etc
- [ ] 5. Create a pipeline to execute the module and load the Target.  Bronze() is a pipeline so all you need to do is check the Overwatch scope to see if sqlAnalytics is present -- if it is, execute the process defined in Step 4, otherwise skip
which leads to step 0-ish which is likely add sqlAnalytics scope to the Overwatch Scope enumeration object and validate any requisite dependencies (there may not be any in bronze)

## What has been completed so far?
Currently, completed the task 1 & 2. Looking for an initial check to see if I am heading in the right direction. I am not sure if the target definition is correct or not .

I will wait for your feedback and then start with Task 3

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/env/Workspace.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/env/Workspace.scala
@@ -53,6 +53,24 @@ class Workspace(config: Config) extends SparkSessionWrapper {
 
   }
 
+  // The below function is supposed to fetch data about DatabricksSQLEndPoints
+  def getSqlEndPointsDF: DataFrame = {
+
+    val jobsEndpoint = "sql/endpoints"
+
+    try {
+      ApiCall(jobsEndpoint, config.apiEnv)
+        .executeGet()
+        .asDF
+        .withColumn("organization_id", lit(config.organizationId))
+    } catch {
+      case e: Throwable => {
+        logger.log(Level.ERROR, "ERROR: Failed to execute jobs/list API call.", e)
+        spark.sql("select ERROR")
+      }
+    }
+
+  }
   /**
    * Exposed config as a public getter to enable access to config for testing. This should not be public facing
    * public function.
--- a/src/main/scala/com/databricks/labs/overwatch/env/Workspace.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/env/Workspace.scala
@@ -54,7 +54,7 @@ class Workspace(config: Config) extends SparkSessionWrapper {
   }
 
   // The below function is supposed to fetch data about DatabricksSQLEndPoints
-  def getSqlEndPointsDF: DataFrame = {
+  def getDatabricksSqlDF: DataFrame = {
 
     val jobsEndpoint = "sql/endpoints"
 
--- a/src/main/scala/com/databricks/labs/overwatch/pipeline/PipelineTargets.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/pipeline/PipelineTargets.scala
@@ -31,6 +31,15 @@ abstract class PipelineTargets(config: Config) {
       partitionBy = Seq("organization_id")
     )
 
+    lazy private[overwatch] val databricksSqlTarget: PipelineTable = PipelineTable(
+      name = "databricks_sql_snapshot_bronze",
+      _keys = Array("id", "Overwatch_RunID"),
+      config,
+      incrementalColumns = Array("Pipeline_SnapTS"),
+      statsColumns = "created_time, creator_user_name, id, Pipeline_SnapTS, Overwatch_RunID".split(", "),
+      partitionBy = Seq("organization_id")
+    )
+
     lazy private[overwatch] val clustersSnapshotTarget: PipelineTable = PipelineTable(
       name = "clusters_snapshot_bronze",
       _keys = Array("cluster_id", "Overwatch_RunID"),


src/main/scala/com/databricks/labs/overwatch/env/Workspace.scala
```diff
@@ -53,6 +53,24 @@ class Workspace(config: Config) extends SparkSessionWrapper {
...
+
+    try {
+      ApiCall(jobsEndpoint, config.apiEnv)
+        .executeGet()
+        .asDF
+        .withColumn("organization_id", lit(config.organizationId))
+    } catch {
+      case e: Throwable => {
+        logger.log(Level.ERROR, "ERROR: Failed to execute jobs/list API call.", e)
+        spark.sql("select ERROR")
```
GeekSheikh, line 69: the latest PR made some changes to the pattern here, see if you can incorporate those.

src/main/scala/com/databricks/labs/overwatch/env/Workspace.scala
```diff
@@ -53,6 +53,24 @@ class Workspace(config: Config) extends SparkSessionWrapper {
 
   }
 
+  // The below function is supposed to fetch data about DatabricksSQLEndPoints
+  def getDatabricksSqlDF: DataFrame = {
+
+    val jobsEndpoint = "sql/endpoints"
```
GeekSheikh, line 59: This is ONE of the endpoints. The key here is to figure out which data is ONLY available via the API calls and which data can otherwise be derived from the auditLogs. We want to use the audit logs where possible and fallback on API calls only where we have to.

GeekSheikh: One of the most important endpoints is going to be the `"sql/history/queries"` and we actually need to pass in a query as well since there's a hidden query to get the metrics. The query string for the get request is `include_metrics=true`. 

Note: The APICall logic already enables passing queries, you just need to pass in the query in the right format.

GeekSheikh: There are also some important differences in the Global SQL Endpoints API. I am almost certain we'll need a couple of snapshot tables for endpoints and one or two appenders to capture all of the SQL endpoint data in bronze.

src/main/scala/com/databricks/labs/overwatch/pipeline/PipelineTargets.scala
```diff
@@ -31,6 +31,15 @@ abstract class PipelineTargets(config: Config) {
       partitionBy = Seq("organization_id")
     )
 
+    lazy private[overwatch] val databricksSqlTarget: PipelineTable = PipelineTable(
+      name = "databricks_sql_snapshot_bronze",
```
GeekSheikh, line 35: As mentioned above, we'll need to have a few different bronze tables, so be diligent on the names as they are hard to change later. Once you become more familiar with the APIs and the data structures, you'll be able to nail this. You'll create one `PipelineTable` per bronze table.

src/main/scala/com/databricks/labs/overwatch/pipeline/PipelineTargets.scala
```diff
@@ -31,6 +31,15 @@ abstract class PipelineTargets(config: Config) {
...
+      _keys = Array("id", "Overwatch_RunID"),
+      config,
+      incrementalColumns = Array("Pipeline_SnapTS"),
+      statsColumns = "created_time, creator_user_name, id, Pipeline_SnapTS, Overwatch_RunID".split(", "),
+      partitionBy = Seq("organization_id")
+    )
```
GeekSheikh, line 41: as you more clearly define which table goes where, just be diligent on making sure the various columns are accurate. It becomes very important later and it wills ave you a lot of headaches. From what I see in the structure, several of those stats columns do not exist

GeekSheikh (commented):
Looks like a good start. I think you need to get more familiar with the data structures before building this out but we do KNOW that there will be a /sql/endpoint list as a snapshot table so I think this is the right place to start to get your feet wet.

Please rebase your branch off of 050 staged in PR #156. It should be merging to main tomorrow (fingers crossed)

GeekSheikh:
closing this PR as it's been reassigned and completed (in progress) by @aman-db

Issue #154, opened 2021-07-06
Data Model - Job, Cluster Improve SCDs


GeekSheikh:
The data model is confusing because the gold tables below more closely relate to the audit logs rather than actual, slow changing objects. Enhance these object definitions so that they are more complete from day 1 and are better defined in the data model.

- job
- cluster

GeekSheikh:
Splitting out Pools and notebooks to complete them at a later date.

GeekSheikh:
While the SCDs are very critical for historical ties, creating a type-1 table/view of the dimension tables to get a simplified picture of the "right now" may simplify this for some users.

GeekSheikh:
planning to keep the slow changing dim settings for now but will explore renaming the SCD tables in gold and creating type-1 tables to give current state of entity

PR #183, opened 2021-07-26
Job dim improvements

### LINKED ISSUE #146: Consumer Data Model - job.new_settings -- Make struct
job.new_settings is a json_string in the consumer model, this should be converted to a struct to simplify querying.

### LINKED ISSUE #166: Add function for getTarget(name: String)
Currently, users must getAllTargets and then use scala to filter for the target with which they want to interact. This should be simpler via api.

GeekSheikh:
Enhanced the job dimension to be more complete and act more as a proper slow-changing dimension with imputed data between events.

Resolves issues with cluster_events to improve the completeness of paginated cluster_events

Schema upgrade will be necessary for jobs dim in silver and gold. Schema upgrade scripts will be included as part of 0.5.1 release.

closes #166 
resolves part of #154 (jobs)
closes #146

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ name := "overwatch"
 
 organization := "com.databricks.labs"
 
-version := "0.5.0.1"
+version := "0.5.1"
 
 scalaVersion := "2.12.12"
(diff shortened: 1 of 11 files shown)


GeekSheikh:
Merged into #243

PR #199, opened 2021-08-16
Issue 196

### LINKED ISSUE #196: Duplicate API_Name (key) in default azure instancedetails
There is a duplicate key in default machineDetails target.

mohanbaabu1996:
closes #196


docs-compose/overwatch_docs/content/FAQ/_index.md
```diff
@@ -0,0 +1,123 @@
+---
```
GeekSheikh, line 1: This file / entire directory `ovewerwatch_docs` should not be begin committed to main. Please remove this directory from your branch. remember when you switch branches, if the file is open it can't remove it.

src/main/resources/Azure_Instance_Details.csv
```diff
@@ -1,89 +1,88 @@
 instance,API_name,vCPUs,Memory_GB,Compute_Contract_Price,On_Demand_Cost_Hourly,Linux_Reserved_Cost_Hourly,Hourly_DBUs
```
GeekSheikh, line 1: Here, it looks like this is a full copy / paste. I cannot assure that only one line has changed, please only remove the one offending line, not copy and paste an entirely new document.

I also see another issue, you have removed a fix I implemented earlier. Notice how on line 2, Memory has moved back from 28.0 to 28. 28 gets interpreted as an Integer and causes issues, we need it to be a double. 

Please just drop the one offending line.

GeekSheikh (changes requested):

PR #208, opened 2021-08-20
Update Azure_Instance_Details.csv

### LINKED ISSUE #206: Update Default Compute Pricing
It seems as though several machine types' default pricing is not accurate in the defaults spreadsheet. This is not an immediate issue as most customers provide their own costs but should be corrected for the next release.

mohanbaabu1996:
closes #206


GeekSheikh (changes requested):
Please do a similar review for the AWS prices

PR #261, opened 2021-10-22
[DRAFT] Filter out the "non-standard" tags coming as default_tags

### LINKED ISSUE #260: Non-standard default_tags should be represented as map, not as struct
In current implementation, the `default_tags` field is treated as struct, but besides the values defined in the [API specification](https://docs.databricks.com/dev-tools/api/latest/clusters.html#get), it could be populated with additional tags, coming from the Azure resource group, or workspace level. And there are situations when these tags could differ in casing, breaking the data processing

Also, should we have a column with all tags combined together, instead of having `custom_tags` & `default_tags` separately?

alexott:
Databricks adds a number of the default tags, like, Creator, ClusterName, ClusterId,
etc. that are returned as `default_tags` objects in the REST API Clusters Get call.  But
this block also returns the tags inherited from the Azure resource group, workspace, etc.
But this block is treated as Spark struct column, and this leads to addition of the
"non-standard" names that are becoming the fields of struct. But in case of mixing the
casing of the same name, the processing is failing when writing data into Delta table.

This fix doing two things:
* leaving only Databricks-specific tags in the `default_tags` column
* converting the original `default_tags` data coming from API into a separate column of
  the map type, so users will still able to query the data using the "non-standard" tags

this fixes #260

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala
@@ -203,6 +203,7 @@ trait BronzeTransforms extends SparkSessionWrapper {
 
     outputDF
       .withColumn("custom_tags", SchemaTools.structToMap(outputDF, "custom_tags"))
+      .withColumn("default_tags", SchemaTools.structToMap(outputDF, "default_tags"))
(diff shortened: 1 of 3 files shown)


alexott:
@GeekSheikh I change `default_tags` to map, but no upgrade procedure yet. Don't know if you already have it, so wanted to wait for your commits

GeekSheikh:
resolved in #277

PR #264, opened 2021-11-05
split DBU and Compute Costs

### LINKED ISSUE #153: instanceDetails - Separate DBU Costs from Machine Costs
Currently instanceDetails table tracks costs for both machine and dbu. This is not the appropriate way to track these costs. These should be broken into two separate tables and the DBU costs should be maintained in one table and the compute costs should be maintained in this table.

All associated cost calculations will need to be altered to reference the new table.

GeekSheikh:
closes #153

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/pipeline/Pipeline.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/pipeline/Pipeline.scala
@@ -118,76 +118,128 @@ class Pipeline(
   }
 
   /**
-   * If latest, active records contain a DBU contract price that differs from those found in the config, return
-   * true, otherwise return false
-   * @param lastRunPricing instanceDetails dataframe with rnk column
+   * TRUE if changes detected
+   * FALSE if no changes
+   * changes mean that the active dbu price for a sku is different in the config than it is in the dbuCosting table
+   * @param lastRunDBUCosts instanceDetails dataframe with rnk column
    * @return
    */
-  private def dbuContractPriceChange(lastRunPricing: DataFrame): Boolean = {
+  private def dbuContractPriceChange(lastRunDBUCosts: DataFrame): Boolean = {
 
-    lastRunPricing
-      .filter('rnk === 1 && 'activeUntil.isNull) // most recent, active records
+    !lastRunDBUCosts
+      .filter('rnk === 1 && 'activeUntil.isNull && 'isActive) // most recent, active records
       .filter(
-        'interactiveDBUPrice === config.contractInteractiveDBUPrice &&
-          'automatedDBUPrice === config.contractAutomatedDBUPrice &&
-          'sqlComputeDBUPrice === config.contractSQLComputeDBUPrice &&
-          'jobsLightDBUPrice === config.contractJobsLightDBUPrice
+        when('sku === "interactive", 'contract_price =!= config.contractInteractiveDBUPrice)
+        .when('sku === "automated", 'contract_price =!= config.contractAutomatedDBUPrice)
+        .when('sku === "sqlCompute", 'contract_price =!= config.contractSQLComputeDBUPrice)
+        .when('sku === "jobsLight", 'contract_price =!= config.contractJobsLightDBUPrice)
       )
       .isEmpty
 
   }
 
   /**
-   * Check for changes between last instance details contract DBU Prices and configured prices from Overwatch
-   * config. If changes are detected, rebuild the slow-changing dim appropriately.
-   * @param cloudDetailTarget PipelineTable for instanceDetails
+   * Check for changes in configured contract DBU Prices.
+   * If changes are detected, rebuild the slow-changing dim appropriately.
+   * @param dbuCostDetail PipelineTable for dbuContractCosts
    */
-  private def updateDBUContractPricing(cloudDetailTarget: PipelineTable): Unit = {
-    val lastRunDBUPriceW = Window.partitionBy('API_Name).orderBy('Pipeline_SnapTS.desc)
-    val lastRunPricing = cloudDetailTarget.asDF()
+  private def updateDBUCosts(dbuCostDetail: PipelineTable): Unit = {
+    val keyCols = dbuCostDetail.keys.map(col)
+    val lastRunDBUPriceW = Window.partitionBy(keyCols: _*).orderBy('Pipeline_SnapTS.desc)
+    val dbuCostDetailDF = dbuCostDetail.asDF()
+
+    val lastRunDBUCosts = dbuCostDetailDF
       .withColumn("rnk", rank().over(lastRunDBUPriceW))
 
-    if (dbuContractPriceChange(lastRunPricing)) {
+    val activeRecord = 'rnk === 1 && 'activeUntil.isNull && 'isActive
+    if (dbuContractPriceChange(lastRunDBUCosts)) {
       val msg = "DBU Pricing differences detected in config. Updating prices."
       logger.log(Level.INFO, msg)
       if (config.debugFlag) println(msg)
 
-      // All active records with new DBU pricing
-      val newDBUPriceRecords = lastRunPricing
-        .filter('rnk === 1 && 'activeUntil.isNull)
-        .withColumn("interactiveDBUPrice", lit(config.contractInteractiveDBUPrice))
-        .withColumn("automatedDBUPrice", lit(config.contractAutomatedDBUPrice))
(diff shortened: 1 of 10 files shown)


src/main/scala/com/databricks/labs/overwatch/pipeline/Pipeline.scala
```diff
@@ -118,76 +118,128 @@ class Pipeline(
...
+        .withColumn("isActive", lit(false))
+
+      // historical records ++ new pricing records ++ newly expired pricing records ++ unchanged active records
+      val updatedDBUCostDetail = newDBUPriceRecords
+        .unionByName(expiringRecords)
+        .unionByName(unchangedActiveCosts)
+        .unionByName(originalExpiredRecords)
         .drop("rnk")
-        .repartition(4)
+        .coalesce(1)
```
alexott, line 195: Why 1? If to minimize number of files, I would rather use optimized writes

src/main/scala/com/databricks/labs/overwatch/pipeline/Pipeline.scala
```diff
@@ -118,76 +118,128 @@ class Pipeline(
...
+    if (config.databaseName != config.consumerDatabaseName) BronzeTargets.dbuCostDetailViewTarget.publish("*")
+  }
+
+  private def publishInstanceDetails(): Unit = {
+    val logMsg = "instanceDetails does not exist and/or does not contain data for this workspace. BUILDING/APPENDING"
+    logger.log(Level.INFO, logMsg)
+    if (getConfig.debugFlag) println(logMsg)
+    val instanceDetailsDF = config.cloudProvider match {
+      case "aws" =>
+        InitializerFunctions.loadLocalCSVResource(spark, "/AWS_Instance_Details.csv")
```
alexott, line 226: just general though - maybe we should host these CSV files somewhere? Or even take directly from Github, so people won't rely onto deployment of the new jars

src/main/scala/com/databricks/labs/overwatch/pipeline/PipelineFunctions.scala
```diff
@@ -338,6 +372,14 @@ object PipelineFunctions {
       s"modules include ${pipelineModules.map(m => s"\n(${m.moduleId}, ${m.moduleName})").mkString("\n")}"))
   }
 
+  private[overwatch] def deriveSKU(isAutomated: Column, sparkVersion: Column): Column = {
+    val isJobsLight = sparkVersion.like("apache_spark_%")
+    when(isAutomated && isJobsLight, "jobsLight")
+      .when(isAutomated && !isJobsLight, "automated")
```
alexott, line 378: just leave `isAutomated`, we already checked for `isJobsLight`

src/main/scala/com/databricks/labs/overwatch/pipeline/PipelineFunctions.scala
```diff
@@ -338,6 +372,14 @@ object PipelineFunctions {
...
+      .when(!isAutomated, "interactive")
```
alexott, line 379: what about Databricks SQL?

src/main/scala/com/databricks/labs/overwatch/utils/Structures.scala
```diff
@@ -145,6 +145,15 @@ case class SimplifiedModuleStatusReport(
 
 case class IncrementalFilter(cronField: StructField, low: Column, high: Column)
 
+case class DBUCostDetail(
+                          organization_id: String,
+                          sku: String,
+                          contract_price: Double,
+                          activeFrom: LocalDate,
+                          activeUntil: Option[LocalDate],
```
alexott, line 153: does it make sense to use LocalDate? I would prefer to use GMT, especially for setups with multiple regions involved

src/main/scala/com/databricks/labs/overwatch/utils/Structures.scala
```diff
@@ -194,47 +205,64 @@ private[overwatch] class ApiCallFailure(
...
-          lower(trim('API_Name)).alias("API_Name"),
-          'rnk, 'rn, 'previousUntil,
-          datediff('activeFrom, 'previousUntil).alias("daysBetweenCurrentAndPrevious")
-        ),
-      Seq("API_Name")
+      invalidReportDF,
+      target.keys
     )
-    .orderBy('API_Name, 'activeFrom, 'activeUntil)
+    .orderBy(target.keys.map(col) ++ Array(col(fromCol), col(untilCol)): _*)
```
alexott, line 231: I see that `target.keys.map(col)` in multiple places. I would just add the function to `PipelineTable` to return seq of columns, something like

```scala
def keyColumns: keys.map(col)
```

src/main/scala/com/databricks/labs/overwatch/utils/Structures.scala
```diff
@@ -194,47 +205,64 @@ private[overwatch] class ApiCallFailure(
...
+    s"A total of $badRecords records were found in conflict. " +
+    s"There cannot be any gaps or overlaps in time for a specific key such that " +
+    s"datediff($fromCol --> $untilCol) returns 0 for all keys. Please correct the data in the table " +
+    s"before continuing with Gold Pipeline."
   throw new BadConfigException(errMsg)
 }
 
-private[overwatch] class PipelineStateException(s: String, val target: Option[PipelineTable]) extends Exception(s) {}
+private[overwatch] class PipelineStateException(s: String, val target: Option[PipelineTable]) extends Exception(s) {
+  println(s)
```
alexott, line 248: Remove `println` everywhere.

src/main/scala/com/databricks/labs/overwatch/pipeline/Pipeline.scala
GeekSheikh: because this lookup actually only holds about 4 records (maybe a few more over time but always less than 100...it's just sooo tiny.

src/main/scala/com/databricks/labs/overwatch/pipeline/Pipeline.scala
GeekSheikh: they are in the resources folder. I get your point but I don't have a great place to store them and make them programmatically available. Now that repo is public, yes, we could stream over https...just haven't done that. Kind of really low on my list. Also, these are just for first-run inits...customers will customize these for their costs so it's mainly for the structure and getting them the 80% of the way.

src/main/scala/com/databricks/labs/overwatch/pipeline/PipelineFunctions.scala
GeekSheikh: fair enough, added to later branch. quicker / easier than fixing here due to pending commits and what not.

src/main/scala/com/databricks/labs/overwatch/pipeline/PipelineFunctions.scala
GeekSheikh: Yep it was in there but no longer belongs here since the entire DBSQL has move over to endpoints now. So the logic is no longer necessary here.

src/main/scala/com/databricks/labs/overwatch/utils/Structures.scala
GeekSheikh: It's a fair thought and I went round and round with this in the early days. There are subtle differences in the way spark and UNIX interpret time so I had to fall back to localDate to get server time to jive with spark time. I think the right way is global TZ of UTC and use of java.sql.timestamp but either way, definitely cannot make these changes until test suite is present.

src/main/scala/com/databricks/labs/overwatch/utils/Structures.scala
GeekSheikh: good idea, let's add that as a feature request.

src/main/scala/com/databricks/labs/overwatch/utils/Structures.scala
GeekSheikh: why? without it the clean error message I've worked hard to construct doesn't appear in the STDOUT of the pipeline. These errors often don't hard-fail the pipeline but rather soft-fail the module...so yeah, they'd be in the log4j but the stdout is much friendlier. I want users to know when they're getting an error, especially in a notebook because that's primarily when they are setting up their first deployment.

GeekSheikh:
thanks for the thorough review. Very much appreciated.

alexott (approved):

PR #267, opened 2021-11-11
first attempt on improving terraform templates

alexott:


terraform/azure/objects/terraform.tfvars
```diff
@@ -1,3 +1,4 @@
-resource_group = "alexott-rg"
-region = "West Europe"
-name_prefix = "overwatch"
+resource_group = "some-rg"
+region = "UK South"
+name_prefix = "overwatch2"
+kv_name = "some-kv2"
```
GeekSheikh, line 4: is there no way to pull these in automatically from the vars. I noticed that these values were hardcoded in other files as well.

GeekSheikh (approved):
looks good but I still think there are some opportunities. Not sure if you wanna combine them in this PR or just do another one later. Putting them below so we'll have them.

Add the following examples
- Provision storage account
- mount storage account
- create cluster policy
- overwatch run job with main class (to have both examples notebook + main class)
- ensure the job owner has access to the key vault key

terraform/azure/objects/terraform.tfvars
alexott: this is an example of the file. Really, `.tfvars` files shouldn't be included into code

alexott:
yep, will do followup PR for the rest of the things

PR #268, opened 2021-11-11
255 init gold tables

### LINKED ISSUE #255: gold table initialization
Currently, when audit logs don't receive edits, there are no events in the audit logs thus there are no records resulting in missing values in the gold table. 

Tables include Pool, job, cluster

Potential Solution
On first run, identify missing keys in the gold table and initialize the gold table from the bronze snapshot. Subsequent edits / updates will flow into the table correctly upon event.

GeekSheikh:
closes #255


alexott (approved):
looks good

PR #269, opened 2021-11-16
enable custom workspace names

### LINKED ISSUE #234: orgId may be not be intuitive for multiple workspace overwatch users
In multiple workspace setups the orgId is used to segregate the data between multiple workspaces. Many users actually know the workspace name / vanity URL for their workspaces but may not use the orgId frequently.

@psg2 mentioned that if perhaps in the gold tables or in the consumption database views we could have an additional column to add the workspace name / vanity URL, in addition to the orgId.

Does that make sense?

`dbutils.notebook.getContext.tags("browserHostName")` would give us the vanity URL. i.e. [demo.cloud.databricks.com](url)

But couldn't find exactly in the APIs or context how to get the actual workspace name.

Potential options would be:

- Creating CASE WHENs in the view definition, could be challenging as each instance will not know about other workspaces in the multi-workspace setup;
- Enriching only the gold tables with the new column to avoid too much redundant data in bronze/gold.

Shall we discuss if this could be an interesting addition?

### LINKED ISSUE #262: PVC - CFT redeploy changes workspace ID
The method through which the workspace ID is derived returns the ID of the load balancer in PVC. The ELB/ILB IDs change when the platform is redeployed. This results in a data disconnect when platform upgrades are completed. There seems to be no static, canonical ID available for a PVC workspace.

Potential Solution -- Specify an optional config override to name the workspace. This would also satisfy the request from Issue #234 as it would allow customers to custom name their workspaces. 

For Azure, this obfuscation could simplify a mapping from custom workspace ID to EH Resource ID to satisfy requirements from Issue #204

GeekSheikh:
closes #262 
closes #234 

Allows users to customize the name of their workspace through the Overwatch config. A new column will appear in each Overwatch table called `Workspace_Name` which will mirror the name. If not configured it defaults to organization_id.

Still need to add support for PVC to switch organization_id to this ... looking into best way to do that now.

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/env/Database.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/env/Database.scala
@@ -22,11 +22,15 @@ class Database(config: Config) extends SparkSessionWrapper {
     this
   }
 
-  private def registerTarget(table: PipelineTable): Unit = {
-    if (!table.exists(catalogValidation = true) && table.exists(pathValidation = true)) {
-      val createStatement = s"create table ${table.tableFullName} " +
-        s"USING DELTA location '${table.tableLocation}'"
-      val logMessage = s"CREATING TABLE: ${table.tableFullName} at ${table.tableLocation}\n$createStatement\n\n"
+  /**
+   * register an Overwatch target table in the configured Overwatch deployment
+   * @param target Pipeline table (i.e. target) as per the Overwatch deployed config
+   */
+  def registerTarget(target: PipelineTable): Unit = {
+    if (!target.exists(catalogValidation = true) && target.exists(pathValidation = true)) {
+      val createStatement = s"create table ${target.tableFullName} " +
+        s"USING DELTA location '${target.tableLocation}'"
+      val logMessage = s"CREATING TABLE: ${target.tableFullName} at ${target.tableLocation}\n$createStatement\n\n"
       logger.log(Level.INFO, logMessage)
       if (config.debugFlag) println(logMessage)
       spark.sql(createStatement)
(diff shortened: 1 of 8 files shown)


GeekSheikh:
Allows users to customize the name of their workspace through the Overwatch config.

Still need to add support for PVC to switch organization_id to this ... looking into best way to do that now.

src/main/scala/com/databricks/labs/overwatch/env/Database.scala
```diff
@@ -129,6 +133,7 @@ class Database(config: Config) extends SparkSessionWrapper {
     var finalDF: DataFrame = df
     finalDF = if (target.withCreateDate) finalDF.withColumn("Pipeline_SnapTS", pipelineSnapTime) else finalDF
     finalDF = if (target.withOverwatchRunID) finalDF.withColumn("Overwatch_RunID", lit(config.runID)) else finalDF
+    finalDF = if (target.withWorkspaceFriendlyName) finalDF.withColumn("Workspace_Name", lit(config.workspaceFriendlyName)) else finalDF
```
alexott, line 136: How it will be migrated for existing data?

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
```diff
@@ -861,33 +861,33 @@ trait GoldTransforms extends SparkSessionWrapper {
...
-      |organization_id, spark_context_id, job_id, job_group_id, execution_id, stage_ids, cluster_id, notebook_id, notebook_path,
+      |organization_id, workspace_name, spark_context_id, job_id, job_group_id, execution_id, stage_ids, cluster_id, notebook_id, notebook_path,
       |db_job_id, db_id_in_job, db_job_type, unixTimeMS, timestamp, date, job_runtime, job_result, event_log_start,
       |event_log_end, user_email
       |""".stripMargin
 
   protected val sparkStageViewColumnMapping: String =
     """
-      |organization_id, spark_context_id, stage_id, stage_attempt_id, cluster_id, unixTimeMS, timestamp, date, stage_runtime,
+      |organization_id, workspace_name, spark_context_id, stage_id, stage_attempt_id, cluster_id, unixTimeMS, timestamp, date, stage_runtime,
```
alexott, line 871: maybe it would be easier to have a lookup table & have view over the existing data?

alexott (approved):
lgtm, just small comments

src/main/scala/com/databricks/labs/overwatch/env/Database.scala
GeekSheikh: during the upgrade script

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
GeekSheikh: these are the views that select over the existing data. I thought about putting it into a lookup but seemed unnecessary for a single value per workspace. This is just exposing it in the presentation layer (which are built views)

PR #272, opened 2021-11-18
externalize optimize option

### LINKED ISSUE #265: Allow for optimize/zorder pipeline to be disabled and add direct entry point from main
Overwatch automatically optimizes and zorders all the tables appropriately as part of the pipeline. For large customers, this can take quite some time and without autoscaling it can eat a lot of hours in the day. 

Adding an entry point into the main class to call the optimize jobs directly will allow user to control schedule and minimize the variability of job runtimes for normal pipeline ingestion.

GeekSheikh:
closes #265

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/pipeline/Bronze.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/pipeline/Bronze.scala
@@ -179,7 +179,7 @@ class Bronze(_workspace: Workspace, _database: Database, _config: Config)
 
     executeModules()
 
-    initiatePostProcessing()
+    if (!config.externalizeOptimize) initiatePostProcessing()
     this
 
   }
--- a/src/main/scala/com/databricks/labs/overwatch/pipeline/PostProcessor.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/pipeline/PostProcessor.scala
@@ -1,28 +1,145 @@
 package com.databricks.labs.overwatch.pipeline
 
-import com.databricks.labs.overwatch.utils.Helpers
+import com.databricks.labs.overwatch.Optimizer
+import com.databricks.labs.overwatch.utils.{Config, Helpers}
+import io.delta.tables.DeltaTable
 import org.apache.log4j.{Level, Logger}
+import org.apache.spark.sql.functions.lit
 
 import scala.collection.mutable.ArrayBuffer
+import scala.collection.parallel.ForkJoinTaskSupport
+import scala.concurrent.forkjoin.ForkJoinPool
 
-class PostProcessor {
+class PostProcessor(config: Config) extends PipelineTargets(config) {
 
   private val logger: Logger = Logger.getLogger(this.getClass)
   private val tablesToOptimize: ArrayBuffer[PipelineTable] = ArrayBuffer[PipelineTable]()
 
+  /**
+   * Specify pipeline target to be optimized
+   * Optimize is not done immediately but later so the cluster can be sized correctly and optimize and zorder ops can
+   * be completed in parallel
+   * @param table Pipeline Target to be marked for optimized
+   */
   private[overwatch] def markOptimize(table: PipelineTable): Unit = {
     tablesToOptimize.append(table)
   }
 
-  def optimize(pipeline: Pipeline, optimizeScaleCoefficient: Int): Unit = {
-    logger.log(Level.INFO, s"BEGINNING OPTIMIZE & ZORDER. " +
-      s"Tables in scope are: ${tablesToOptimize.map(_.tableFullName).mkString(",")}")
-
-    PipelineFunctions.scaleCluster(pipeline, optimizeScaleCoefficient)
-    // TODO -- spark_events_bronze -- put in proper rules -- hot fix due to optimization issues
-    Helpers.parOptimize(tablesToOptimize.toArray.filterNot(_.name == "spark_events_bronze"), maxFileSizeMB = 128)
-    Helpers.parOptimize(tablesToOptimize.toArray.filter(_.name == "spark_events_bronze"), maxFileSizeMB = 32)
+  /**
+   * Several module output targets have a few other, intermediate "targets" that aren't the output of an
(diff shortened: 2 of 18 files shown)


src/main/scala/com/databricks/labs/overwatch/pipeline/PostProcessor.scala
```diff
@@ -1,28 +1,145 @@
...
+         |MERGE CONDITION: $mergeCondition
+         |UPDATE EXPRESSION: $updateExpr
+         |""".stripMargin
+
+    if (config.debugFlag) println(msg)
+    logger.log(Level.INFO, msg)
+    DeltaTable.forPath(pipelineStateTarget.tableLocation).alias("target")
+      .merge(lastOptimizedUpdateDF, mergeCondition)
+      .whenMatched
+      .updateExpr(updateExpr)
```
alexott, line 116: what will happen when we have new tables added into list to optimize?

src/main/scala/com/databricks/labs/overwatch/utils/Structures.scala
```diff
@@ -124,13 +128,15 @@ case class ModuleStatusReport(
       status,
       recordsAppended,
       lastOptimizedTS,
-      vacuumRetentionHours
+      vacuumRetentionHours,
```
alexott, line 131: it also makes sense to externalize vacuum - we don't need big cluster for it, so running it inside the same job cluster is just waste of resources.

src/main/scala/com/databricks/labs/overwatch/utils/Tools.scala
```diff
@@ -306,14 +306,14 @@ object Helpers extends SparkSessionWrapper {
     tablesPar.foreach(tbl => {
       try {
         val zorderColumns = if (tbl.zOrderBy.nonEmpty) s"ZORDER BY (${tbl.zOrderBy.mkString(", ")})" else ""
-        val sql = s"""optimize ${tbl.tableFullName} $zorderColumns"""
-        println(s"optimizing: ${tbl.tableFullName} --> $sql")
+        val sql = s"""optimize delta.`${tbl.tableLocation}` $zorderColumns"""
```
alexott, line 309: maybe optimize each organization_id separately if necessary? Otherwise we'll rewrite a lot of data in the multi-workspace setup

src/main/scala/com/databricks/labs/overwatch/utils/Tools.scala
```diff
@@ -122,7 +122,7 @@ object Helpers extends SparkSessionWrapper {
    * TODO: rename to defaultParallelism
    */
   private def parallelism: Int = {
-    Math.min(driverCores, 8)
+    Math.max(driverCores * 2, 8)
```
alexott, line 125: we should be careful with running parallel optimize, because it may overload cluster. VACUUM is ok with parallelization

alexott (commented):
I'm thinking if it makes sense to add generated column, like, timestamp rounded to week or month, so we won't re-optimize old data again & again.

src/main/scala/com/databricks/labs/overwatch/pipeline/Bronze.scala
```diff
@@ -179,7 +179,7 @@ class Bronze(_workspace: Workspace, _database: Database, _config: Config)
 
     executeModules()
 
-    initiatePostProcessing()
+    if (!config.externalizeOptimize) initiatePostProcessing()
```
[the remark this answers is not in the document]
GeekSheikh: nope, much more efficient to have one global externalize option. Will be in docs. :)

src/main/scala/com/databricks/labs/overwatch/pipeline/PostProcessor.scala
GeekSheikh: If they are targets that result from a module, there will be another record in the pipReport. If future tables are supporting tables they will be added to the optimize of the parent target they support. Either way this merge should continue to work as expected as far as I can tell, did I miss something?

src/main/scala/com/databricks/labs/overwatch/utils/Structures.scala
GeekSheikh: I assumed folks would just use an autoscaling cluster for this which would handle it right? This was one of the primary reasons for externalizing optimize was to enable a standard autoscaling cluster to be efficient.

src/main/scala/com/databricks/labs/overwatch/utils/Tools.scala
GeekSheikh: optimize and zorder are done within partitions, right? Since every target / table is partitioned by organization_id all optimization workloads should be confined to the partition directory[ies], right?

src/main/scala/com/databricks/labs/overwatch/utils/Tools.scala
GeekSheikh: good call, I will reduce parallelize with optimize. I have it set in a way so as not to overload it but I will make cluster-size recommendations in the docs for this. Optimize is generally safe at driverCores - 1 from my experience.

GeekSheikh:
Thanks for all the comments. 🥇

src/main/scala/com/databricks/labs/overwatch/utils/Tools.scala
alexott: Yes, but if people will run it from multiple workspaces, it will run for all partitions

alexott (approved):

src/main/scala/com/databricks/labs/overwatch/utils/Tools.scala
GeekSheikh: right, I will try to make that clear in the docs. If I limit it to a single org_id here then users will be REQUIRED to run this on every workspace which is not what I want them to do.

GeekSheikh:
> I'm thinking if it makes sense to add generated column, like, timestamp rounded to week or month, so we won't re-optimize old data again & again.

Optimize already handles this. Optimizations are placed into supersets. Once the set is full, no more optimizes will go against that until the next superset has enough to fill the next superset.

PR #273, opened 2021-11-19
add gold stream entity preview

### LINKED ISSUE #168: Streams - Add Streams Detail Entity to Gold Data Model
The streamsDF has been created and just needs to be tested and added to Gold data model

### LINKED ISSUE #274: Single Node Clusters -- Null Costs
When calculating compute costs for single node clusters, the worker type and details are null and thus result in a null cost. 

Looks like already handled for worker dbu costs

when cluster_type == "single_node" set worker compute cost to 0

GeekSheikh:
closes #168 
closes #274 

Adding spark streams table to gold as preview entity. Need more feedback from customers so this entity will likely evolve a bit through the 0.6.x lineage.


alexott (approved):
lgtm

PR #276, opened 2021-11-23
edge case bug fix for nested null structs

GeekSheikh:
Sometimes structFromJson would be unable to derive a schema (i.e. when all records were null). Enforced minimum schemas where this problem is expected to occur.


alexott (approved):
lgtm

PR #277, opened 2021-11-24
upgrade 060 script plus helpers

### LINKED ISSUE #260: Non-standard default_tags should be represented as map, not as struct

GeekSheikh:
closes #260


src/main/scala/com/databricks/labs/overwatch/pipeline/Initializer.scala
```diff
@@ -164,8 +164,8 @@ class Initializer(config: Config) extends SparkSessionWrapper {
       config.setIsPVC(true)
       if (config.debugFlag) println(pvcDetectedMsg)
       logger.log(Level.INFO, pvcDetectedMsg)
-      require(config.workspaceFriendlyName.toLowerCase != config.organizationId.toLowerCase,
-        s"PVC workspaces require the 'workspaceFriendlyName' to be configured in the Overwatch configs for " +
+      require(config.workspaceName.toLowerCase != config.organizationId.toLowerCase,
+        s"PVC workspaces require the 'workspaceName' to be configured in the Overwatch configs for " +
```
alexott, line 168: but it's named `workspace_name` in the RawConfig. Should we use it here? Or you're referring to the field in `Config`?

src/main/scala/com/databricks/labs/overwatch/ParamDeserializer.scala
```diff
@@ -158,7 +158,7 @@ class ParamDeserializer() extends StdDeserializer[OverwatchParams](classOf[Overw
       getOptionDouble(masterNode, "intelligentScaling.coeff").getOrElse(1.0)
     )
 
-    val workspaceFriendlyName = getOptionString(masterNode, "workspaceFriendlyName")
+    val workspace_name = getOptionString(masterNode, "workspace_name")
```
alexott, line 161: why `workspace_name` not `workspaceName` for variables, etc.?

src/main/scala/com/databricks/labs/overwatch/utils/Upgrade.scala
```diff
@@ -266,4 +274,570 @@ object Upgrade extends SparkSessionWrapper {
...
+        s"prior to upgrade stored as Dataframe at $initialTableVersionsPath"
+
+      logger.log(Level.INFO, failMsg)
+      println(failMsg)
+
+      upgradeStatus.toSeq.toDF().write.format("delta").mode("overwrite")
+        .save(upgradeReportPath)
+
+      initialTableVersions.toSeq.toDF().write.format("delta").mode("overwrite")
+        .save(initialTableVersionsPath)
```
alexott, line 296: will this work in case if we already modified the schema of the table(s)? Maybe we should use `overwriteSchema` ?

src/main/scala/com/databricks/labs/overwatch/pipeline/Initializer.scala
GeekSheikh: yeah had to change this around, it's working now. I wanted this to be camelCase in the config but in the database tables i wanted it to match the format as all the other columns `workspace_name`...took some doing but finally got it to work. Only down-side is that the OverwatchParams case class has a weird `workspace_name` field in it but it's ok. It caused issues with the upgrades to do it other ways due to state.

src/main/scala/com/databricks/labs/overwatch/ParamDeserializer.scala
GeekSheikh: for the reason I said above, I want it to be in all the tables as "workspace_name". this is the only column in OverwatcParams that winds up moving through the other tables. I can work cleaning up the pattern after the upgrade or something but it's challenging to change it here due to the way the state of the pipeline works.

src/main/scala/com/databricks/labs/overwatch/utils/Upgrade.scala
GeekSheikh: there's only one write to this path per upgrade init. It's suffixed with the timestamp to the millisecond. This means that if someone needed to resume later they could and then could reference these values each time they kicked off the upgrade.

src/main/scala/com/databricks/labs/overwatch/utils/SchemaTools.scala
```diff
@@ -123,10 +123,7 @@ object SchemaTools extends SparkSessionWrapper {
       println(s"WARNING: The json schema for column $c was not parsed correctly, please review.")
     }
     if (jsonSchema.isEmpty) {
-      if (minSchema.isEmpty) logger.log(Level.WARN, s"A schema cannot be inferred for $c give the data " +
-        s"in this dataframe. A minimum schema may be provided but hasn't been. Will implicitly cast $c to a null of " +
-        s"StringType. If errors persist, please review this field.")
-      lit(null).cast(minSchema.getOrElse(StringType))
+      lit(null)
```
alexott, line 126: that's dangerous - Delta just silently drop null columns, so when you read data back you may not have it. maybe use `lit(null).cast("struct<>")` if we expect to return a struct column?

src/main/scala/com/databricks/labs/overwatch/env/Workspace.scala
```diff
@@ -127,7 +132,74 @@ class Workspace(config: Config) extends SparkSessionWrapper {
...
+
+  /**
+   * get EXISTING dataset[s] metadata within the configured Overwatch workspace
+   * @return Seq[WorkspaceDataset]
+   */
+  def getWorkspaceDatasets: Seq[WorkspaceDataset] = {
+    dbutils.fs.ls(config.etlDataPathPrefix).map(dataset => {
+      val path = dataset.path
+      val uri = Helpers.getURI(path)
+      val name = if(dataset.name.endsWith("/")) dataset.name.dropRight(1) else dataset.name
```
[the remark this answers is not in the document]
GeekSheikh: It's a fair point but that's a completely unsupported functionality (dropping raw files into the Overwatch ETL dir)...nonetheless people do weird stuff all the time. 

I've added the filter for the fileStatus
`.filter(_.isDir).map(dataset => ...)`

src/main/scala/com/databricks/labs/overwatch/pipeline/Pipeline.scala
```diff
@@ -367,31 +368,33 @@ class Pipeline(
 
   private[overwatch] def initiatePostProcessing(): Unit = {
 
-    postProcessor.optimize(this, 12)
+    if (!config.externalizeOptimize) postProcessor.optimize(this, 12)
```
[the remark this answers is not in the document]
GeekSheikh: Only used in this one place for the pos-processor but moved to constant

src/main/scala/com/databricks/labs/overwatch/utils/SchemaTools.scala
GeekSheikh: That is exactly the behavior I want. If the dynamic schema does have data for the column, drop it. 
Example:  Col(x) has a string value of "{}" (empty json) there is no data thus the source dataframe will not include that column at all. If it's a required column downstream, the `df.verifyMinimumSchema(schema)` will add it back and type it correctly.

src/main/scala/com/databricks/labs/overwatch/env/Database.scala
```diff
@@ -157,6 +157,7 @@ class Database(config: Config) extends SparkSessionWrapper {
            |MERGE CONDITION: $mergeCondition
            |""".stripMargin
       logger.log(Level.INFO, mergeDetailMsg)
+      spark.conf.set("spark.databricks.delta.commitInfo.userMetadata", config.runID)
       // TODO -- when DBR 9.1 LTS GA, use LSM (low-shuffle-merge) to improve pipeline
```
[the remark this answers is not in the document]
alexott: I'm talking about LSM, maybe just say people that they can enable it. I would wait until GA to have it enabled by default

alexott (approved):
lgtm

src/main/scala/com/databricks/labs/overwatch/env/Database.scala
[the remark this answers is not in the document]
GeekSheikh: right it's not enabled by default

PR #280, opened 2021-12-08
handle corrupted audit logs for DataSourceId

### LINKED ISSUE #278: Corrupted Audit logs breaking audit log loads in some cases
Databricks recently published new actions to the audit logs wherein both `dataSourceId` and `DataSourceId` (and two other columns) are in the top-level schema. This causes there to be duplicate columns (since case-sensitivity is off by default) and thus breaking schema inference for audit logs. 

This will only happen in certain workspaces with certain use cases in DBSQL while the audit log publications were corrupted.

Solution, add try/catch around this block to catch the specific exception and enable case-sensitivity temporarily when this error occurs.

This is not an Overwatch bug but a case where Overwatch was not designed to handle corrupted schemas. 

The error you will see is
```
AnalysisException: Found duplicate column(s) in the data schema: `<column_name`>
```

The there columns that could result in this are

- DataSourceId
- AlertId
- DashboardId

GeekSheikh:
closes #278

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ name := "overwatch"
 
 organization := "com.databricks.labs"
 
-version := "0.5.0.6"
+version := "0.5.0.6.1"
 
 scalaVersion := "2.12.12"
 scalacOptions ++= Seq("-Xmax-classfile-name", "78")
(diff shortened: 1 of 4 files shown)


alexott (approved):
lgtm

alexott:
btw, why not just 0.5.07?

GeekSheikh:
> btw, why not just 0.5.07?

I would but there are 0 features, it's a hot-fix.

PR #293, opened 2022-01-05
v06_bugZaps

### LINKED ISSUE #288: instance_pools - API not returning expected structure -- missing column instance_pools
This error occurs when the pools scope is enabled but there are no instance pools. This will be converted to a soft failure in next patch release, in the meantime, this can be resolved by removing `pools` from the scopes list when there are no instance_pools in the workspace.

Instead of Scopes: all us Scopes: `audit,sparkEvents,jobs,clusters,clusterEvents,notebooks,accounts`

related to #286

### LINKED ISSUE #289: num_workers column missing when no autoscaling clusters present in workspace
When a workspace has no autoscaling clusters, there's an error in Silver_ClusterStateDetail module.

This will be resolved in next patch release. In the meantime, this issue can be resolved by simply creating an autoscaling cluster, starting it, and then stopping it and rerunning the pipeline.

Relates to #286

### LINKED ISSUE #291: Missing valid regex character for EH Connection String SharedAccessKey
`+` is a valid base64 character which should be permitted in the regex at 

https://github.com/databrickslabs/overwatch/blob/3fb6414a8a813a5652dce9fcd256162f78202086/src/main/scala/com/databricks/labs/overwatch/pipeline/PipelineFunctions.scala#L36

### LINKED ISSUE #292: ClusterStateDetailSilver - cluster_size.autoscale.num_workers doesn't exist
This field can be missing when there are no autoscaling clusters. This can also occur in a slightly different manner when all clusters are autoscaling. 

Error is a result of this column derivation:
https://github.com/databrickslabs/overwatch/blob/3fb6414a8a813a5652dce9fcd256162f78202086/src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala#L899

Proposed solution:
Add to minimum schema requirements data contract for input DF cluster_events_bronze.

### LINKED ISSUE #294: joinWithLag - can produce duplicate records
When joining with lagged df and the incremental data contains the same data as the lag data, there's a possibility of duplicates in the df. 

Proposed Solution: drop produced duplicate keys from the join

### LINKED ISSUE #295: cluster cloud (vm) costs attributed when cluster is terminated
DQ bug found -- clusters in terminated state could have costs attributed during upgrades to v6 and/or during historical loads.

### LINKED ISSUE #296: spark_confs - special characters - SchemaTools
Some field names in schemas can consist of ONLY special characters. In such cases all field name characters will be removed resulting in a struct that cannot be converted to a map since the column name (map key) will be null string.

Example: someone tries to comment out a spark_conf like:
```
# spark.sql.shuffle.partitions 200
```
When the `#` gets converted to a nullString the resulting map key will be "" causing an error. 

Proposed Solution: Instead of converting special characters to a null string, convert them to a `_`.

GeekSheikh:
closes #289 
closes #292
closes #291 
closes #288 
closes #294 
closes #295 
closes #296

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
@@ -783,6 +783,27 @@ trait GoldTransforms extends SparkSessionWrapper {
     df.select(sparkExecutionCols: _*)
   }
 
+  private def cleanseDesc(colName: String): Column = {
+    trim(
+      when(col(colName).like("%<br/>%"), regexp_replace(col(colName), "<br/>", " "))
+        .when(col(colName).like("%\n%"), regexp_replace(col(colName), "\\n", " "))
+        .otherwise(col(colName))
+    )
+  }
+
+  private def deriveStreamDetails(streamArrayCol: Column): Column = {
+    val idPos = array_position(streamArrayCol, "id") + 1
+    val runIdPos = array_position(streamArrayCol, "runId") + 1
+    val batchIdPos = when(array_contains(streamArrayCol, "batchId"), array_position(streamArrayCol, "batchId"))
+      .when(array_contains(streamArrayCol, "batch"), array_position(streamArrayCol, "batch"))
+      .otherwise(lit(0)) + 1
+    struct(
+      streamArrayCol(idPos).alias("stream_id"),
+      streamArrayCol(runIdPos).alias("stream_run_id"),
+      streamArrayCol(batchIdPos).alias("stream_batch_id")
+    )
+  }
+
   protected def buildSparkStream(
                                   sparkStreamsGoldTarget: PipelineTable,
                                   sparkExecutionsSilverDFLag30d: DataFrame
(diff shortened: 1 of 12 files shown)


src/main/scala/com/databricks/labs/overwatch/pipeline/PipelineFunctions.scala
```diff
@@ -33,7 +33,7 @@ object PipelineFunctions {
       case None =>
         connectionString
     }
-    if (!retrievedConnectionString.matches("^Endpoint=sb://[a-zA-Z0-9-_:;/.=]*;SharedAccessKey=[a-zA-Z0-9-_:;/.=]*$")) {
+    if (!retrievedConnectionString.matches("^Endpoint=sb://[a-zA-Z0-9-_:;/.=]*;SharedAccessKey=[a-zA-Z0-9-_:;/.=+]*$")) {
```
alexott, line 36: we can potentially simplify regex to `^Endpoint=sb://.*;SharedAccessKey=.*$` to avoid additional validation problems

src/main/scala/com/databricks/labs/overwatch/utils/SchemaTools.scala
```diff
@@ -211,7 +211,7 @@ object SchemaTools extends SparkSessionWrapper {
    * @return
    */
   private def sanitizeFieldName(s: String): String = {
-    s.replaceAll("[^a-zA-Z0-9_]", "")
+    s.replaceAll("[^a-zA-Z0-9_]", "_")
```
alexott, line 214: my main concern here is that we may have two names of the same length that will be converted into `___` strings of the same length.

alexott (approved):
overall looks ok. there are some comments

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
```diff
@@ -783,6 +783,27 @@ trait GoldTransforms extends SparkSessionWrapper {
     df.select(sparkExecutionCols: _*)
   }
 
+  private def cleanseDesc(colName: String): Column = {
+    trim(
+      when(col(colName).like("%<br/>%"), regexp_replace(col(colName), "<br/>", " "))
+        .when(col(colName).like("%\n%"), regexp_replace(col(colName), "\\n", " "))
+        .otherwise(col(colName))
```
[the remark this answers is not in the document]
GeekSheikh: I mainly did it to make it easier to read. I don't think it will actually pass through the data twice, it's a case statement it will either match case a or case b or else. Nonetheless, after thinking about it I think I like your idea better. Please review the updated logic and make sure they look correct.

src/main/scala/com/databricks/labs/overwatch/pipeline/PipelineFunctions.scala
GeekSheikh: done, was considering doing this. thanks

src/main/scala/com/databricks/labs/overwatch/utils/SchemaTools.scala
GeekSheikh: that is possible but this is always wrapped in scrubSchema which, in the case of a duplicate there will be a unique suffix pasted. If you have ideas on how to improve this, I'm all ears. :) 

see [generateUniques](https://github.com/databrickslabs/overwatch/blob/3fb6414a8a813a5652dce9fcd256162f78202086/src/main/scala/com/databricks/labs/overwatch/utils/SchemaTools.scala#L236-L266) which is part of scrubSchema pipeline along with private function `sanitizeFields`

PR #298, opened 2022-01-31
0603 upgrade improvements

### LINKED ISSUE #297: Regression - Underscores Appearing in Bronze Fields
Issue #296 required a change to the schema scrubber logic such that special characters would be switched to an underscore instead of a null string. There are several bronze fields that rely on this null string stripping to cleanse column names with spaces.

### LINKED ISSUE #299: Azure Audit Log Loading - Minimum Rows
When landing Azure audit logs via streams the code is currently set to continue to stream until there are 0 records processed in a batch. When streaming in Azure audit logs on busy workspaces, it's possible that the stream never "completes" because rows are getting into the queue faster than the batches are processing, thus never allowing azure to complete a load. This can result in a stream never completing.

Proposal - allow a configurable "minimumEventsPerTrigger" and default it to 10. It's currently hard-coded to 0.

GeekSheikh:
closes #297 
closes #299

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ name := "overwatch"
 
 organization := "com.databricks.labs"
 
-version := "0.6.0.2"
+version := "0.6.0.3"
 
 scalaVersion := "2.12.12"
 scalacOptions ++= Seq("-Xmax-classfile-name", "78")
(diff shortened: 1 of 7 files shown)


alexott (approved):
lgtm

PR #315, opened 2022-02-28
pools bug fix

GeekSheikh:


src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
```diff
@@ -515,8 +515,12 @@ trait SilverTransforms extends SparkSessionWrapper {
       .selectExpr("*", "requestParams.*").drop("requestParams", "Overwatch_RunID")
 
     val poolsRawWithStructs = poolsRaw
-      .withColumn("aws_attributes", SchemaTools.structFromJson(spark, poolsRaw, "aws_attributes"))
-      .withColumn("azure_attributes", SchemaTools.structFromJson(spark, poolsRaw, "azure_attributes"))
+      .withColumn("aws_attributes",
+        when('aws_attributes.isNull, struct(lit(null).cast("string").alias("emptyKey")))
```
alexott, line 519: maybe we can use existing fields instead of introducing the new fields? `availability` is good candidate - it exists in all clouds

alexott (approved):

PR #322, opened 2022-03-10
Upgrade bug fixed

### LINKED ISSUE #321: Azure - Upgrade 0506 -> 060[3,4] directly -- STEP 10: error rebuilding silver targets
The issue is that a new parameter was introduced into the config in version 0603 and was not properly added to the upgrade path causing a failure.

GeekSheikh:
closes #321

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/utils/Upgrade.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/utils/Upgrade.scala
@@ -310,16 +310,8 @@ object Upgrade extends SparkSessionWrapper {
    * @param pipelineReportPath pass in database name
    * @return
    */
-  private def getLatestWorkspaceByOrg(pipelineReportPath: String): Array[OrgConfigDetail] = {
+  def getLatestWorkspaceByOrg(pipelineReportPath: String): Array[OrgConfigDetail] = {
     val oldestDateByModuleW = Window.partitionBy('organization_id).orderBy('Pipeline_SnapTS.desc)
-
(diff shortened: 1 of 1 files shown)


src/main/scala/com/databricks/labs/overwatch/utils/Upgrade.scala
```diff
@@ -310,29 +310,32 @@ object Upgrade extends SparkSessionWrapper {
...
-    )
-
     val overwatchParamsCols = Array("auditLogConfig", "tokenSecret", "dataTarget", "badRecordsPath", "overwatchScope",
       "maxDaysToLoad", "databricksContractPrices", "primordialDateString", "intelligentScaling", "workspace_name",
       "externalizeOptimize")
 
     // Necessary because if recovering pipeline upgrade the columns may or may not exist so both scenarios
     // must be handled. This ensures both columns are present and present only once
-    val orgRunDetailsBase = spark.read.format("delta").load(pipelineReportPath)
+    val orgRunDetailsBase = spark.read.format("delta").load("/mnt/overwatch_global/05x/global_share/pipeline_report")
```
alexott, line 322: Is it correct to hardcore path?

alexott (changes requested):

src/main/scala/com/databricks/labs/overwatch/utils/Upgrade.scala
GeekSheikh: nope that's a booboo. Customer noticed too lol dangit.

alexott (approved):

PR #379, opened 2022-05-19
PE-939: AWS Secrets Manager

dyon78:
Add possibility to work with AWS Secrets Manager


.github/workflows/publish-artifacts.yml
```diff
@@ -0,0 +1,24 @@
+name: Publish
```
alexott, line 1: I think that this file should be removed

dyon78: Removed

dyon78:
@GeekSheikh could you review please?

PR #381, opened 2022-05-19
Ov 38 use new api

### LINKED ISSUE #87: Improve ApiCall.scala
We still need to use open source API helpers but this code got pretty messy in development. There are many challenges with apis and escape characters using DBConnect which have differences between mac/linux/windows. I believe this could be GREATLY simplified but didn't have the time to work it out.

### LINKED ISSUE #90: Module 1005 - API Custom Batching - Tests
API batching for Module 1005, Bronze_ClusterEventLogs. The reason this is such an important component is because the ONLY place to get this information (which drives cost and cluster util / everything) is the API. Some customers have 100s of thousand cluster event states per day. This is too much data to pull to the driver all at once, thus I batch it (not talking API pagination, that's working) but I built a custom batcher to query number of cluster events between x and y dates and then batch them to 500K. Pull 500K events, write to table, repeat. This data expires after 60d (so I'm told) and once the data expires in the API, it's not accessible programmatically anywhere else again.

Historical loads and large customers will hit this upper bound and it's critical to ensure the process is working. I envision several tests. I believe unit tests can be used to simulate a small upper bound (set 20 as the upper bound to batch for testing) and ensure that the data is pulled, written and then continues to pull the next batch.

This test isn't as important as the higher-level tests you're working on now as errors here would likely present themselves later but in some cases they may not. I wanted to put this on the radar as a required bronze test set need.

Relevant code is below.
The clusterEvents API call requires the clusterID, thus a list of all clusterIDs with new events must be created to send to the API calls as part of the API query. The section below is the section responsible for ensuring all clusters with new events are accounted for, this is a VERY CRITICAL component of Overwatch.
https://github.com/databrickslabs/overwatch/blob/4e3daa9e7e54722f3b8692b0cd2e39703e41a4df/src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala#L297-L330

Batching Code:
https://github.com/databrickslabs/overwatch/blob/4e3daa9e7e54722f3b8692b0cd2e39703e41a4df/src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala#L339-L382

### LINKED ISSUE #290: Allow for proxy headers in API Calls
Thus far proxies have been handled through system and env vars. There are situations wherein this is insufficient. Users must be able to pass in proxy info and auths into Overwatch API calls.

DEPENDS ON #302 

**Proposed Solution:**
Add Proxy Config Parameters as an optional parameter to APIEnv constructor. Whether or not this is to be created from env vars or part of the overwatch config is tbd.

### LINKED ISSUE #302: APICall.scala -- Refactor / Modernization
APICall has been sticthed together since the beginning of Overwatch and greatly needs an overhaul. Databricks has released several new types of paginated APIs that are needed for DBSQL data grabs.

* Modernize & Simplify APICall.scala
* Add full unit testing suite with simulated responses
* Add functionality for Issue #290

### LINKED ISSUE #389: parallelize clusterEvents Bronze Module
driver parellelization should be supported in new APICallV2.

Sriram-databricks:
Updated the framework to use the new API

closes #290 
closes #302 
closes #90 
closes #389 
closes #87


src/main/scala/com/databricks/labs/overwatch/ApiCallV2.scala
```diff
@@ -0,0 +1,515 @@
...
+      .setApiName(apiName)
+  }
+
+  def apply(apiEnv: ApiEnv, apiName: String, queryJsonString: String) = {
+    new ApiCallV2(apiEnv)
+      .setApiName(apiName)
+      .setQuery(queryJsonString)
+  }
+
+  def apply(apiEnv: ApiEnv, apiName: String, queryJsonArray: Array[String],tempSuccessPath:String,errorTempPath:String) = {
```
GeekSheikh, line 31: if we externalize executeBatch as mentioned below, do we still need this apply signature? If not, let's externalize and drop this signature and the steTempLocation funcs

src/main/scala/com/databricks/labs/overwatch/ApiMeta.scala
```diff
@@ -0,0 +1,98 @@
...
+ */
+trait ApiMeta {
+
+  private val logger: Logger = Logger.getLogger(this.getClass)
+  var paginationKey:String="";
+  var paginationToken:String="";
+  var dataframeColumns:String="";
+  var apiCallType:String="";
+  var microBatchSize = 50
+  var tempLocation = "/local_disk0/OverWatch"
```
GeekSheikh, line 15: change this default to _ ?

src/main/scala/com/databricks/labs/overwatch/ApiMeta.scala
```diff
@@ -0,0 +1,98 @@
...
+ * Configuration for each API.
+ */
+trait ApiMeta {
+
+  private val logger: Logger = Logger.getLogger(this.getClass)
+  var paginationKey:String="";
+  var paginationToken:String="";
+  var dataframeColumns:String="";
+  var apiCallType:String="";
+  var microBatchSize = 50
```
GeekSheikh, line 14: This should be controlled from ApiEnv and further externalized to the global overwatch config so that it can be easily overridden from config.

src/main/scala/com/databricks/labs/overwatch/ApiMeta.scala
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * Configuration for each API.
+ */
+trait ApiMeta {
+
+  private val logger: Logger = Logger.getLogger(this.getClass)
+  var paginationKey:String="";
+  var paginationToken:String="";
+  var dataframeColumns:String="";
+  var apiCallType:String="";
```
GeekSheikh, line 13: perfect example, don't default this to something that will never make sense...either initialize it to null `_` or something that is "normal" like "GET"

src/main/scala/com/databricks/labs/overwatch/ApiMeta.scala
```diff
@@ -0,0 +1,98 @@
+package com.databricks.labs.overwatch
+import org.apache.log4j.{Level, Logger}
+
+/**
+ * Configuration for each API.
+ */
+trait ApiMeta {
```
GeekSheikh, line 7: let's add an init log function to this trait and ensure it gets called each time a class that extends this trait is instantiated. This will ensure that any instantiation that extends this will log the details of the entire meta defaults and overrides.

src/main/scala/com/databricks/labs/overwatch/env/Workspace.scala
```diff
@@ -1,7 +1,7 @@
 package com.databricks.labs.overwatch.env
 
 import com.databricks.dbutils_v1.DBUtilsHolder.dbutils
-import com.databricks.labs.overwatch.ApiCall
+import com.databricks.labs.overwatch.{ApiCall, ApiCallV2}
```
GeekSheikh, line 4:
```suggestion
import com.databricks.labs.overwatch.{ApiCallV2}
```

We should be able to remove ApiCall from the imports once V2 is fully implemented.

src/main/scala/com/databricks/labs/overwatch/env/Workspace.scala
```diff
@@ -127,10 +122,7 @@ class Workspace(config: Config) extends SparkSessionWrapper {
...
   def getWorkspaceUsersDF: DataFrame = {
     val workspaceEndpoint = "workspace/list"
-    ApiCall(workspaceEndpoint, config.apiEnv, Some(Map("path" -> "/Users")), debugFlag = config.debugFlag)
-      .executeGet()
-      .asDF
-      .withColumn("organization_id", lit(config.organizationId))
+    ApiCallV2(config.apiEnv,workspaceEndpoint).execute().asDF().withColumn("organization_id", lit(config.organizationId))
   }
 
   private def clusterState(apiEnv: ApiEnv): String = {
```
GeekSheikh, line 128: Still Using ApiCall v1, update this func to use v2

src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala
```diff
@@ -1,6 +1,6 @@
 package com.databricks.labs.overwatch.pipeline
```
GeekSheikh, line 1: Will review this file after we externalize `executeBatch` from ApiCallV2

src/main/scala/com/databricks/labs/overwatch/utils/Config.scala
```diff
@@ -355,7 +355,7 @@ class Config() {
       if (!rawToken.matches("^(dapi|dkea)[a-zA-Z0-9-]*$")) throw new BadConfigException(s"contents of secret " +
         s"at scope:key $scope:$key is not in a valid format. Please validate the contents of your secret. It must be " +
         s"a user access token. It should start with 'dapi' ")
-      setApiEnv(ApiEnv(isLocalTesting, workspaceURL, rawToken, packageVersion))
+      setApiEnv(ApiEnv(isLocalTesting, workspaceURL, rawToken, packageVersion,1,1,runID))
```
GeekSheikh, line 358: why are we force setting batch sizes to 1 here and then later setting it to 50 inside APIMeta. These should be externalized into the Overwatch global config and defaulted to something that makes sense, like 50.

src/main/scala/com/databricks/labs/overwatch/utils/Structures.scala
```diff
@@ -33,7 +34,21 @@ case class DatabricksContractPrices(
                                      jobsLightDBUCostUSD: Double = 0.10
                                    )
 
-case class ApiEnv(isLocal: Boolean, workspaceURL: String, rawToken: String, packageVersion: String)
+case class ApiEnv(isLocal: Boolean, workspaceURL: String, rawToken: String, packageVersion: String,successBatchSize:Int=0,errorBatchSize:Int=0,runID:String="")
```
GeekSheikh, line 37: A 3rd set of defaults for batch sizes -- these should all be defaulted to the same, reasonable value.

Curious as to if we can set the package version as a function inside the case class so that it can be automatic. So remove `packageVersion` from the signature completely and create a function called `packageVersion` that get's the package version in the same way as in the config.scala

build.sbt
```diff
@@ -15,6 +15,11 @@ libraryDependencies += "com.databricks" % "dbutils-api_2.12" % "0.0.5" % Provide
 libraryDependencies += "com.amazonaws" % "aws-java-sdk-s3" % "1.11.595" % Provided
 libraryDependencies += "io.delta" % "delta-core_2.12" % "1.0.0" % Provided
 libraryDependencies += "org.scalaj" %% "scalaj-http" % "2.4.2"
+//For API Modernization
+libraryDependencies += "com.tdunning" % "json" % "1.8"
+libraryDependencies += "log4j" % "log4j" % "1.2.17"
```
GeekSheikh, line 20: where is this being used? why do we need this?

build.sbt
```diff
@@ -15,6 +15,11 @@ libraryDependencies += "com.databricks" % "dbutils-api_2.12" % "0.0.5" % Provide
 libraryDependencies += "com.amazonaws" % "aws-java-sdk-s3" % "1.11.595" % Provided
 libraryDependencies += "io.delta" % "delta-core_2.12" % "1.0.0" % Provided
 libraryDependencies += "org.scalaj" %% "scalaj-http" % "2.4.2"
+//For API Modernization
+libraryDependencies += "com.tdunning" % "json" % "1.8"
```
GeekSheikh, line 19: see comments throughout code. I'm very sensitive to adding new dependencies but am happy to do so where justified. We just need to make sure this is what we want to use as a team.

src/main/scala/com/databricks/labs/overwatch/ApiCallV2.scala
```diff
@@ -0,0 +1,515 @@
...
+  private var debugFlag = false //Debug flag to print the information when required. d
+  private var allowUnsafeSSL = false //Flag to make the unsafe ssl.
+  private var jsonKey = "" //Key name for pagination.
+  private var jsonValue = "" //Key value for pagination.
+  private val readTimeoutMS = 60000 //Read timeout.
+  private val connTimeoutMS = 10000 //Connection timeout.
+  private var errorTempPath:String=""
+  private var queryJsonArray:Array[String]=null
+  private var successBatchSize:Int=0
+  private var errorBatchSize:Int=0
```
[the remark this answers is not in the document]
Sriram-databricks: We don't need this 2 variable anymore because we will get the batch size from APIEnv.So removing it.

src/main/scala/com/databricks/labs/overwatch/ApiCallV2.scala
```diff
@@ -0,0 +1,515 @@
...
+    this
+  }
+ def setSuccessBatchSize(successBatchSize:Int):this.type ={
+    this.successBatchSize=successBatchSize
+    this
+  }
+  def setErrorBatchSize(errorBatchSize:Int):this.type ={
+    this.errorBatchSize=errorBatchSize
+    this
+  }
```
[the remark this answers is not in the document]
Sriram-databricks: Done

src/main/scala/com/databricks/labs/overwatch/ApiCallV2.scala
```diff
@@ -0,0 +1,515 @@
...
+
+  /**
+   * Setting up the debug flag which can be useful to print certain things.
+   * @param value
+   * @return
+   */
+  private def setDebugFlag(value: Boolean): this.type = {
+    debugFlag = value
+    this
+  }
```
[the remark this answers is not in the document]
Sriram-databricks: removed

src/main/scala/com/databricks/labs/overwatch/ApiCallV2.scala
```diff
@@ -0,0 +1,515 @@
...
+      jsonKey = jsonObject.keySet().iterator().next().toString
+      jsonValue = jsonObject.getString(jsonKey)
+    }catch {
+      case e:Exception=>{
+        val excMsg = "Got the exception while parsing provided query json "
+        logger.log(Level.WARN, excMsg + e.getMessage)
+        throw new Exception(e)
+      }
+    }
+  }
```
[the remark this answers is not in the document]
Sriram-databricks: Created a new function in JsonUtil without using external libs.

src/main/scala/com/databricks/labs/overwatch/ApiCallV2.scala
```diff
@@ -0,0 +1,515 @@
...
+
+  /**
+   *Hibernate in-case of too many API call request.
+   * @param response
+   */
+  def hibernate( response: HttpResponse[String]): Unit = {
```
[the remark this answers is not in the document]
Sriram-databricks: done

src/main/scala/com/databricks/labs/overwatch/utils/Config.scala
Sriram-databricks: done

src/main/scala/com/databricks/labs/overwatch/utils/Structures.scala
Sriram-databricks: We don't have any use of packageVersion inside API call, i guess we can remove it from APIEnv case class.

build.sbt
Sriram-databricks: removed

build.sbt
Sriram-databricks: removed the libs

build.sbt
```diff
@@ -15,6 +15,11 @@ libraryDependencies += "com.databricks" % "dbutils-api_2.12" % "0.0.5" % Provide
...
+libraryDependencies += "log4j" % "log4j" % "1.2.17"
+libraryDependencies += "org.powermock" % "powermock-module-junit4" % "2.0.2" % Test
```
GeekSheikh: This is weird -- I see the add here but when I pull the branch I don't see this dependency...trying to figure out why it's different.

src/main/scala/com/databricks/labs/overwatch/ApiCallV2.scala
```diff
@@ -0,0 +1,515 @@
...
+
+  }
+
+  /**
+   * Check if the response contains the key for next page and creates the required variable which can be used to create the API call for next page.
+   * @param jsonObject response as jsonObject received from API call.
+   */
+  def paginate(jsonObject: JSONObject): Unit = {
+    if (jsonObject.keySet().contains(apiMeta.paginationKey)) { //Checking for pagination
+      if (endPoint.equals("sql/history/queries")) { //For sql/history/queries api we have a different mechanism to get the next page token
```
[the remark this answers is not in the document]
GeekSheikh: ?? I thought we moved this to the ApiMeta, no?

src/main/scala/com/databricks/labs/overwatch/ApiCallV2.scala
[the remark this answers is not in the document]
Sriram-databricks: @GeekSheikh jsontoMap still exits but the implementation has been changed as below.I have added a utility function in JsonUtils and using that one and it doesn’t uses any external library .
/**
   * Parse the json string and get the data in the form of key and value.
   */
  private def jsonToMap: Unit = {
    try {
      val (jsonKey, jsonValue) = JsonUtils.getJsonKeyValue(jsonQuery)
      this.jsonKey = jsonKey
      this.jsonValue = jsonValue
    } catch {
      case e: Exception => {
        val excMsg = "Got the exception while parsing provided query json "
        logger.log(Level.WARN, excMsg + e.getMessage)
        throw new Exception(e)
      }
    }
  }

src/main/scala/com/databricks/labs/overwatch/ApiCallV2.scala
```diff
@@ -0,0 +1,515 @@
...
+   */
+  def jsonToMap: Unit = {
+    try {
+      val jsonObject = new JSONObject(jsonQuery)
+      jsonKey = jsonObject.keySet().iterator().next().toString
+      jsonValue = jsonObject.getString(jsonKey)
+    }catch {
+      case e:Exception=>{
+        val excMsg = "Got the exception while parsing provided query json "
+        logger.log(Level.WARN, excMsg + e.getMessage)
```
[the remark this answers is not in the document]
Sriram-databricks: code modified

src/main/scala/com/databricks/labs/overwatch/ApiCallV2.scala
```diff
@@ -0,0 +1,515 @@
...
+  private var allowUnsafeSSL = false //Flag to make the unsafe ssl.
+  private var jsonKey = "" //Key name for pagination.
+  private var jsonValue = "" //Key value for pagination.
+  private val readTimeoutMS = 60000 //Read timeout.
+  private val connTimeoutMS = 10000 //Connection timeout.
+  private var errorTempPath:String=""
+  private var queryJsonArray:Array[String]=null
+  private var successBatchSize:Int=0
+  private var errorBatchSize:Int=0
+  private var errorArray:util.ArrayList[ApiErrorDetail] = new util.ArrayList[ApiErrorDetail]();
```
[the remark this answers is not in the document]
GeekSheikh: @Sriram-databricks -- is this done?

src/main/scala/com/databricks/labs/overwatch/ApiCallV2.scala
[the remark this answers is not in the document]
GeekSheikh: @Sriram-databricks -- you fixed this one but please fix the rest in the code. Anywhere we catch or rethrow an error, please log the error like you did here and like noted above.

src/main/scala/com/databricks/labs/overwatch/ApiCallV2.scala
```diff
@@ -0,0 +1,515 @@
...
+       |""".stripMargin
+  }
+
+
+  /**
+   * Get the required columns from the received dataframe.
+   * @param rawDF The raw dataframe which is received as part of API response.
+   * @return Dataframe which contains the required columns.
+   */
+  def processRawData(rawDF: DataFrame): DataFrame = {
```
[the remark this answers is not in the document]
GeekSheikh: Whoops -- that was my type -- I meant `extrapolateSupportedStructure`

src/main/scala/com/databricks/labs/overwatch/ApiCallV2.scala
```diff
@@ -0,0 +1,515 @@
...
+    val resultDFFieldNames = rawDF.schema.fieldNames
+    if (!resultDFFieldNames.contains(apiMeta.dataframeColumns) && apiMeta.dataframeColumns != "*") { // if known API but return column doesn't exist
```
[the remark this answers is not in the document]
GeekSheikh: suggest changing name from `dataframeColumns` to `dataframeColumn` since it's a string and only accepts a single column. If more than one column were to be passed in map logic and an array would need to be used. Unnecessary for now but maybe needed in future. Wording indicates that this may already be possible.

src/main/scala/com/databricks/labs/overwatch/ApiCallV2.scala
```diff
@@ -0,0 +1,515 @@
...
+   * Function responsible for making the API request.
+   * @return
+   */
+  def execute(): this.type = {
+    try {
+      val response = getResponse
+      val responseMapper = parseJson(response)
+      responseCodeHandler(response)
+      apiResponseArray.put(responseMapper.rawJsonObject)
+      if (apiMeta.storeInTempLocation) {
```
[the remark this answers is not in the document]
GeekSheikh: still see `storeInTempLocation` in ApiMeta -- not a problem just want to understand this

src/main/scala/com/databricks/labs/overwatch/utils/Structures.scala
GeekSheikh: why are you defaulting batch sizes to 0? I strongly suggest we default this to something reasonable. 0 is prob the worst default as it will throw strange errors

src/main/scala/com/databricks/labs/overwatch/ApiCallV2.scala
```diff
@@ -0,0 +1,549 @@
...
+  private val connTimeoutMS = 10000 //Connection timeout.
+  private var _printFlag: Boolean = true
+
+  def endPoint: String = _endPoint
+
+  def jsonQuery: String = _jsonQuery
+
+  def serverBusyCount: Int = _serverBusyCount
+
+  def printFlag: Boolean = _printFlag
```
GeekSheikh, line 69: It seems like some of these getters are useless for a customer and are better off not being exposed. We generally want to minimize the autoComplete options for users. Let's remove all the getters that are very deep internals such as printFlag, jsonKey, jsonValue, apiResponseArray etc...these are only used for internal control and shouldn't be directly accessed by the user from what i understand.

src/main/scala/com/databricks/labs/overwatch/ApiCallV2.scala
```diff
@@ -0,0 +1,549 @@
...
+    this
+  }
+
+  /**
+   * Setting up the api name and api metadata for that api.
+   *
+   * @param value
+   * @return
+   */
+  private def setApiName(value: String): this.type = {
```
GeekSheikh, line 160: can we rename this? This is the primary entry / build point, the constructor for the class essentially...how about `buildAPI`?

GeekSheikh (commented):
didn't get too far, will need to pick this up later. sorry for the multiple reviews.

src/main/scala/com/databricks/labs/overwatch/ApiMeta.scala
```diff
@@ -0,0 +1,177 @@
...
+  def dataframeColumn: String = _dataframeColumn
+
+  def apiCallType: String = _apiCallType
+
+  def storeInTempLocation: Boolean = _storeInTempLocation
+
+  def apiV: String = _apiV
+
+  def tuplePaginationObject: Boolean = _tuplePaginationObject
+
```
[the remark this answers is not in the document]
Sriram-databricks: done

src/main/scala/com/databricks/labs/overwatch/ApiMeta.scala
```diff
@@ -0,0 +1,177 @@
...
+}
+
+/**
+ * Factory class for api Metadata.
+ */
+class ApiMetaFactory {
+  private val logger: Logger = Logger.getLogger(this.getClass)
+
+  def getApiClass(_apiName: String): ApiMeta = {
```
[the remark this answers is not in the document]
Sriram-databricks: done

src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala
```diff
@@ -454,90 +459,114 @@ trait BronzeTransforms extends SparkSessionWrapper {
...
-    logger.log(Level.INFO, s"NUMBER OF BATCHES: ${clusterEventsBuffer.length} \n" +
-      s"ESTIMATED EVENTS: ${clusterEventsBuffer.length * batchSize.toInt}")
+    logger.log(Level.INFO, "Calling New APIv2, Number of cluster id:" + clusterIDs.size + " run id :" + apiEnv.runID)
+    val tmpClusterEventsSuccessPath = s"$tempWorkingDir/success" + apiEnv.runID
+    val tmpClusterEventsErrorPath = s"$tempWorkingDir/error" + apiEnv.runID
+    val finalResponseCount = clusterIDs.size
+    var responseCounter = 0;
+    var apiResponseArray = Collections.synchronizedList(new util.ArrayList[String]())
+    var apiErrorArray = Collections.synchronizedList(new util.ArrayList[String]())
+    implicit val ec = ExecutionContext.fromExecutor(Executors.newFixedThreadPool(4))
```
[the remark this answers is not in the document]
Sriram-databricks: done

src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala
```diff
@@ -454,90 +459,114 @@ trait BronzeTransforms extends SparkSessionWrapper {
...
 
-    logger.log(Level.INFO, s"NUMBER OF BATCHES: ${clusterEventsBuffer.length} \n" +
-      s"ESTIMATED EVENTS: ${clusterEventsBuffer.length * batchSize.toInt}")
+    logger.log(Level.INFO, "Calling New APIv2, Number of cluster id:" + clusterIDs.size + " run id :" + apiEnv.runID)
+    val tmpClusterEventsSuccessPath = s"$tempWorkingDir/success" + apiEnv.runID
+    val tmpClusterEventsErrorPath = s"$tempWorkingDir/error" + apiEnv.runID
+    val finalResponseCount = clusterIDs.size
+    var responseCounter = 0;
+    var apiResponseArray = Collections.synchronizedList(new util.ArrayList[String]())
+    var apiErrorArray = Collections.synchronizedList(new util.ArrayList[String]())
```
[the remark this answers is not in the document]
Sriram-databricks: done

src/main/scala/com/databricks/labs/overwatch/utils/Config.scala
```diff
@@ -44,6 +44,8 @@ class Config() {
   private var _contractAutomatedDBUPrice: Double = _
   private var _contractSQLComputeDBUPrice: Double = _
   private var _contractJobsLightDBUPrice: Double = _
+  private var _successBatchSize:Int= 50
+  private var _errorBatchSize:Int= 50
```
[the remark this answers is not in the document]
Sriram-databricks: done

Sriram-databricks (commented):
implemented all the comments

src/main/scala/com/databricks/labs/overwatch/ApiCallV2.scala
```diff
@@ -0,0 +1,554 @@
...
+   */
+  def execute(): this.type = {
+    try {
+      val response = getResponse
+      println(response.body)
+      responseCodeHandler(response)
+      _apiResponseArray.add(response.body)
+      paginate(response.body)
+      if (_printFinalStatusFlag) {
+        println(buildSleepDetailMessage())
```
GeekSheikh, line 511: don't print -- just log

src/main/scala/com/databricks/labs/overwatch/ApiCallV2.scala
```diff
@@ -0,0 +1,554 @@
...
+        if (apiEnv.successBatchSize <= _apiResponseArray.size()) { //Checking if its right time to write the batches into persistent storage
+          val responseFlag = Helpers.writeMicroBatchToTempLocation(successTempPath, _apiResponseArray.toString)
+          if (responseFlag) { //Clearing the resultArray in-case of successful write
+            setApiResponseArray(new util.ArrayList[String]())
+          }
+        }
+      }
+      paginate(response.body)
+      if (_printFinalStatusFlag) {
+        println(buildSleepDetailMessage())
```
GeekSheikh, line 468: only print inside the hibernate functions -- not necessary when not hibernating. This is duplicate

src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala
```diff
@@ -434,6 +439,103 @@ trait BronzeTransforms extends SparkSessionWrapper {
...
+                apiErrorArray = Collections.synchronizedList(new util.ArrayList[String]())
+              }
+            }
+
+          }
+          responseCounter = responseCounter + 1
+          logger.log(Level.ERROR, "Future failure message: " + e.getMessage, e)
+      }
+    }
+    while (responseCounter < finalResponseCount) {
```
GeekSheikh, line 486: As I feared, this can lock up the module -- when there are failures in some of the clusters -- such as deleted clusters or other errors -- the responses are never captured so it sleeps forever waiting for those hanging threads to complete.

At the very least -- for now, we must have a timeout and make sure those get sent to error
Better would be to make sure the response accounts for successes + errors
Best (for later) identify better logic control to wait for futures to complete -- as already discussed.

src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala
```diff
@@ -434,6 +439,103 @@ trait BronzeTransforms extends SparkSessionWrapper {
...
+      //As we are using Futures and running 4 threads in parallel, We are checking if all the treads has completed the execution or not.
+      // If we have not received the response from all the threads then we are waiting for 5 seconds and again revalidating the count.
+      println("Sleeping for 5 seconds...Response received count : " + responseCounter + ",Expected Response count : " + finalResponseCount)
```
GeekSheikh, line 489: please change this to logging only unless cumulative sleep time becomes > 120s. There needs to be some timeout or logic to ensure the module cannot get hung forever. Upon timeout the entire module should fail with an appropriate error message that is bubbled up to the error report.

Change the message to `Waiting for other queued API Calls to complete; cumulative wait time <n> seconds`

src/main/scala/com/databricks/labs/overwatch/ApiCallV2.scala
```diff
@@ -0,0 +1,554 @@
...
+
+  /**
+   * Function responsible for making the API request.
+   *
+   * @return
+   */
+  def execute(): this.type = {
+    try {
+      val response = getResponse
+      println(response.body)
```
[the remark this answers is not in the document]
Sriram-databricks: Oops done by mistake

src/main/scala/com/databricks/labs/overwatch/ApiCallV2.scala
```diff
@@ -0,0 +1,554 @@
...
+      .setQuery(queryJsonString)
+  }
+
+  def apply(apiEnv: ApiEnv, apiName: String, queryJsonString: String, tempSuccessPath: String) = {
+    new ApiCallV2(apiEnv)
+      .buildApi(apiName)
+      .setQuery(queryJsonString)
+      .setSuccessTempPath(tempSuccessPath)
+  }
+  def apply(apiEnv: ApiEnv, apiName: String, queryMap: Map[String, String]) = {
```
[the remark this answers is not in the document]
Sriram-databricks: Map[String,Any] is not supported by  Http(s"""${apiEnv.workspaceURL}/${apiMeta.apiV}/${endPoint}""").params it only supports Map[String,String]

src/main/scala/com/databricks/labs/overwatch/ApiCallV2.scala
```diff
@@ -0,0 +1,554 @@
...
+      setPrintFlag(false)
+    }
+    apiMeta.apiCallType match {
+      case "POST" =>
+        response =
+          try {
+            Http(s"""${apiEnv.workspaceURL}/${apiMeta.apiV}/${endPoint}""")
+              .copy(headers = httpHeaders)
+              .postData(jsonQuery)
+              .options(reqOptions)
```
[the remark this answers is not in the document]
Sriram-databricks: for POST request we are using json String as it is not supporting Map<String,String>.
Below example is running fine.

 val jsonQuery = """{"cluster_id": "0511-074343-c456edtd","num_workers":2}"""
  ApiCallV2(apiEnv, endpoint, jsonQuery).execute()

src/main/scala/com/databricks/labs/overwatch/ApiCallV2.scala
Sriram-databricks: done

src/main/scala/com/databricks/labs/overwatch/ApiCallV2.scala
Sriram-databricks: actually the name of the function is confusing so renaming the function as buildDetailMessage(). 
At end of the API call this function will print the stats of the API call like.

API call Endpoint: https://adb-2222170229861029.9.azuredatabricks.net/api/2.0/clusters/resize
queryString: {"cluster_id": "0511-074343-c456edtd","num_workers":2}
Server Busy Count:0
Total Sleep Time:0 seconds
Successful api response count:1

src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala
Sriram-databricks: i guess we will always get a response even its a failure, in case of the cluster is not found it will send an api exception which we take into count as a response.

for now i am implementing timeout of 5 minutes.If we dont get any new response even waiting for 5 mints it will throw an exception.

src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala
Sriram-databricks: done

GeekSheikh (approved):
Great work! :)

PR #428, opened 2022-06-09
perf tuning 

GeekSheikh:
adding some perf optimization for 611 

Relying more heavily on AQE is great idea but still needs a bit of tuning. working through some optimizations

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/pipeline/PipelineFunctions.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/pipeline/PipelineFunctions.scala
@@ -151,6 +151,42 @@ object PipelineFunctions {
       .alias(defaultAlias)
   }
 
+  def optimizeDFForWrite(
+                        df: DataFrame,
+                        target: PipelineTable
+                        ): DataFrame = {
+
+    var mutationDF = df
+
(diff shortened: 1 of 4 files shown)


src/main/scala/com/databricks/labs/overwatch/pipeline/PipelineFunctions.scala
```diff
@@ -151,6 +151,42 @@ object PipelineFunctions {
...
+  def optimizeDFForWrite(
+                        df: DataFrame,
+                        target: PipelineTable
+                        ): DataFrame = {
+
+    var mutationDF = df
+
+    mutationDF = if (target.zOrderBy.nonEmpty) {
+      mutationDF.moveColumnsToFront(target.zOrderBy ++ target.statsColumns)
+    } else mutationDF
```
alexott, line 163: where the data skip columns is set? Or we rely on the default value of 32?

alexott (approved):

src/main/scala/com/databricks/labs/overwatch/pipeline/PipelineFunctions.scala
GeekSheikh: I leave it at default but ensure the keys are at the front.

Issue #433, opened 2022-06-14
[FEAT] -- Add additional snapshots to bronze


GeekSheikh:
Add the following API lists as bronze entities for future consumption in Silver/Gold

- [instance profiles](https://docs.databricks.com/dev-tools/api/latest/instance-profiles.html#list)
- [global init scripts](https://docs.databricks.com/dev-tools/api/latest/global-init-scripts.html#operation/get-scripts)
- [job runs](https://docs.databricks.com/dev-tools/api/latest/jobs.html#operation/JobsRunsList)
  - Use API version 2.1
  - Use APICallV2
  - `expand_tasks=true`
- [tokens](https://docs.databricks.com/dev-tools/api/latest/tokens.html#list)

aman-db:
@GeekSheikh  - In order to start the dev for this issue. Should we create a branch from [OV-38_Use_New_API](https://github.com/databrickslabs/overwatch/tree/OV-38_Use_New_API) instead of 
[431_new_bronze_snaps](https://github.com/databrickslabs/overwatch/tree/431_new_bronze_snaps)? As [OV-38_Use_New_API](https://github.com/databrickslabs/overwatch/tree/OV-38_Use_New_API) contains APICallV2

PR #441, opened 2022-07-06
433 instanceprofile snaps to bronze

aman-db:
partially resolves https://github.com/databrickslabs/overwatch/issues/433
Added instance profile snapshots in cluster scope for bronze.


piyushagarwalla-db (approved):
lgtm

PR #473, opened 2022-08-04
In ApiCallV2 removed the exception and sending empty dataframe

### LINKED ISSUE #461: [BUG] - Error when Pools scope enabled but no pools
The pipeline will error in bronze if the scopes are configured to include pools but there are no pools.

The expected behavior is to return `Empty` and move on but instead it's throwing an exception and breaking the pipeline.

ganeshdogiparthi-db:
closes #461
Negative testing --> With existing code bronze process is failing and terminating without executing the whole scope.
Unit testing --> With modified code bronze process is completed even if there is no pool available.

out of scope
I have not tested the data(tables generated by overwatch) after the code change.

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/env/Workspace.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/env/Workspace.scala
@@ -105,11 +105,21 @@ class Workspace(config: Config) extends SparkSessionWrapper {
         .withColumn("organization_id", lit(config.organizationId))
     }
     catch {
-      case e: ApiCallEmptyResponse =>
-        val message = s"Exceotion in reading pool list ${e}"
+      /**In case if there is no cluster pools available in workspace .asDF() call throws ApiCallEmptyResponse exception.
+       * Handle the ApiCallEmptyResponse exception and move on to the rest of the modules
+       */
+      case emptyResponseError: ApiCallEmptyResponse =>
+        val message = s"Exceotion in reading pool list ${emptyResponseError}"
         println(message)
         logger.error(message)
         spark.emptyDataFrame
+
+      /**
+       * Any other exception other than ApiCallEmptyResponse exception. raise the exception and exit the process
+       */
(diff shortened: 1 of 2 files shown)


src/main/scala/com/databricks/labs/overwatch/env/Workspace.scala
```diff
@@ -98,10 +98,19 @@ class Workspace(config: Config) extends SparkSessionWrapper {
...
-      .asDF()
-      .withColumn("organization_id", lit(config.organizationId))
+    try {
+      ApiCallV2(config.apiEnv, poolsEndpoint)
+        .execute()
+        .asDF()
+        .withColumn("organization_id", lit(config.organizationId))
+    }
+    catch {
+      case e: Exception =>
```
Sriram-databricks, line 108: Put try catch only on  .asDF() method,
 If we put try-catch on .execute() then we might miss other exceptions like no access/invalid URL etc.

ganeshdogiparthi-db: changed the type of exception catching to ApiCallEmptyResponse. This should rule out empty responses from API and raise exceptions in case of any other exception. Please check.

GeekSheikh: agree with @Sriram-databricks -- the right place to put this error handler is in the APICallV2 asDF function to ensure that the errors bubble up to the pipReport. 

If I understand correctly, the actual error is not the API call as that will work fine and return 0 results, but the issue is rather that when we try to convert it to a DF and select the nested column, the col doesn't exist and is causing the breakage. I think we need a generic handler for this in asDF in APICallV2 so that any APICall that is tried to be converted to a DF but is missing the expected column should error.

Sriram-databricks (approved):
Done nicely great work Ganesh and Sourav

GeekSheikh (approved):
lgtm - thanks!

PR #481, opened 2022-08-12
Pipeline function test

sundarshankar89:
Fixes pipeline test for timezone

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ name := "overwatch"
 
 organization := "com.databricks.labs"
 
-version := "0.6.1.1"
+version := "0.6.1.2"
 
 scalaVersion := "2.12.12"
 scalacOptions ++= Seq("-Xmax-classfile-name", "78")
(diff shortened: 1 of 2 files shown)


alexott (dismissed):
One test is still failing:

```
[info] Tests for epochMilliToTs
[info] - should convert epoch time to timestamp and preserve milliseconds *** FAILED ***
[info]   Expected "2022-08-10 [10:5]9:39.388", but got "2022-08-10 [07:2]9:39.388" (PipelineFunctionsTest.scala:246)
[info]   + Given a dataframe with epoch milliseconds time of long type
[info]   + When function is called on this column
[info]   + Then returns a column of type timestamp and preserves milliseconds
```

sundarshankar89:
It should be fixed now.

sundarshankar89:
@alexott Ready for review.

GeekSheikh:
@sundarshankar89 -- I dunno what's going on but I'm seeing 4 tests failing now on branch `pipelineFunctionTest`

GeekSheikh:
Note @sundarshankar89 and @souravbaner-da -- some of the tests pass when run inside of IntelliJ but fail on `sbt clean assembly` please validate both.

GeekSheikh (commented):
Is there a way to limit the scope of disabling parallel execution to just what we need?

souravbaner-da:
> Is there a way to limit the scope of disabling parallel execution to just what we need?

@GeekSheikh It seems parallel execution disabling does not work on specific testcase. But here I have made changes and disable the parallel execution for Test Script only rather than whole project build. All test cases are passing now.

GeekSheikh (approved):

GeekSheikh (approved):

PR #498, opened 2022-08-23
 Add tags to instance pool and default tags to cluster table

### LINKED ISSUE #430: add instance pools tags
Instance pools have tags and I'm not sure they are getting propagated all the way to gold; let's verify we are getting these tags to gold.

### LINKED ISSUE #492: [FEAT] - Add default tags to cluster gold
custom_tags is in gold but default_tags are not getting pulled through.

sudharshanraja-db:
closes #430

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
@@ -5,6 +5,7 @@ import com.databricks.labs.overwatch.utils.{NoNewDataException, SchemaTools, Spa
 import org.apache.log4j.Level
 import org.apache.spark.sql.expressions.Window
 import org.apache.spark.sql.functions._
+import org.apache.spark.sql.types.{MapType, StringType}
 import org.apache.spark.sql.{Column, DataFrame}
 
 trait GoldTransforms extends SparkSessionWrapper {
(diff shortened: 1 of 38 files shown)


src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
```diff
@@ -934,7 +936,7 @@ trait GoldTransforms extends SparkSessionWrapper {
     """
       |organization_id, workspace_name, instance_pool_id, serviceName, timestamp, date, actionName, instance_pool_name, node_type_id,
       |idle_instance_autotermination_minutes, min_idle_instances, max_capacity, preloaded_spark_versions,
-      |azure_attributes, create_details, delete_details, request_details
+      |azure_attributes, create_details, delete_details, request_details,tags
```
GeekSheikh, line 939: these are for end-user views -- the order is important, just sticking this at the end may not be the best place. I moved it after the cloud_attributes columns

GeekSheikh (commented):
This is a preliminary review -- I am going to take a look at the work done for adding default_tags to cluster but this is the review for instance pools. I also made several changes so you can review the commit to see the changes and we can discuss them.

GeekSheikh:
confirmed, default_tags is not published by clusters service either. In fact, it appears that default_tags are not published to audit logs for any services -- confirmed by the snippet below.

```
display(
  table("overwatch_global_etl.audit_log_bronze")
    .filter($"requestParams.default_tags".isNotNull)
    .select('serviceName, 'actionName).distinct
)
```

GeekSheikh (approved):
With a few changes this is ready to go, the result is custom_tags and aws_attributes was added to cluster_gold and custom_tags was added to pools_gold (and their views)

PR #504, opened 2022-08-30
Fix eventhubs dependency

alexott:


alexott:
@GeekSheikh I've tested on both AWS & Azure, and new approach works fine

GeekSheikh (approved):
lgtm -- thanks for digging in and getting this to work without needing the dependency.

PR #510, opened 2022-09-02
#327 issue to handle photon cost in the calculation

### LINKED ISSUE #327: Add custom DBU costing for Photon
in clusterStateFact and jobRunCostPotentialFact, the derived costs should be able to calculate DBU costs for photon usage as well. These DBU Contract Prices for photon should also be customizable in the config as other DBU pricing

ganeshdogiparthi-db:
added runtime_engine in clusters_snapshot_bronze, cluster_spec_silve, cluster_gold tables and added extra calculation for photon-enabled clusters #327 

closes #327

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
@@ -46,7 +46,8 @@ trait GoldTransforms extends SparkSessionWrapper {
       'organization_id,
       'deleted_by,
       'createdBy.alias("created_by"),
-      'lastEditedBy.alias("last_edited_by")
+      'lastEditedBy.alias("last_edited_by"),
+      'runtime_engine
     )
     df.select(clusterCols: _*)
   }
@@ -232,7 +233,7 @@ trait GoldTransforms extends SparkSessionWrapper {
       )
 
     val nodeTypeLookup = clusterSpec.asDF
-      .select('organization_id, 'cluster_id, 'cluster_name, 'custom_tags, 'timestamp, 'driver_node_type_id, 'node_type_id, 'spark_version)
+      .select('organization_id, 'cluster_id, 'cluster_name, 'custom_tags, 'timestamp, 'driver_node_type_id, 'node_type_id, 'spark_version, 'runtime_engine)
       .withColumn("sku", PipelineFunctions.deriveSKU(isAutomated('cluster_name), 'spark_version))
       .toTSDF("timestamp", "organization_id", "sku")
       .lookupWhen(dbuCostDetailsTSDF)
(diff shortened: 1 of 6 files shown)


src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
```diff
@@ -757,10 +757,11 @@ trait SilverTransforms extends SparkSessionWrapper {
           'spark_conf,
           'driver_instance_pool_id,
           'instance_pool_id,
-          'spark_version,
+          'effective_spark_version.alias("spark_version"),
```
GeekSheikh, line 760: all added fields need to be added to the appropriate `minimumSchema` in Schema.scala. Be sure this is in the requestParams section of audit_log_master_schema and that the downstream consumers of these tables have them -- if they are missing and placed in a select it will blow up the pipeline.

This is true for effective_spark_version and runtime_engine (only two I see added ATM.

GeekSheikh, line 760: all added fields need to be added to the appropriate minimumSchema in Schema.scala. Be sure this is in the requestParams section of audit_log_master_schema and that the downstream consumers of these tables have them -- if they are missing and placed in a select it will blow up the pipeline.

This is true for effective_spark_version and runtime_engine (only two I see added ATM.

src/main/scala/com/databricks/labs/overwatch/pipeline/TransformFunctions.scala
```diff
@@ -394,10 +394,21 @@ object TransformFunctions {
...
+      val isPhotonEnabled = upper(runtimeEngine).equalTo("PHOTON")
+      //Check if the cluster is not a SQL warehouse/endpoint
+      val isNotAnSQlWarehouse = !upper(sku).equalTo("SQLCOMPUTE")
+      //This is the default logic for DBU calculation
+      val defaultCalculation = dbu_H * computeTime_H * nodeCount * dbuRate_H * smoothingCol.getOrElse(lit(1))
+
+      //asign the variables and return column with calculation
       coalesce(
-        when(isDatabricksBillable, dbu_H * computeTime_H * nodeCount * dbuRate_H * smoothingCol.getOrElse(lit(1)))
+        when(isDatabricksBillable && isPhotonEnabled && isNotAnSQlWarehouse , defaultCalculation * 2)
```
GeekSheikh, line 410: clean! good work!

GeekSheikh (changes requested):

souravbaner-da (approved):

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
```diff
@@ -289,13 +289,14 @@ trait GoldTransforms extends SparkSessionWrapper {
       // using noise generators to fill with two passes to reduce skew
       .fillMeta(clusterPotMetaToFill, clusterPotKeys, clusterPotIncrementals, noiseBuckets = getTotalCores) // scan back then forward to fill all
 
+
```
GeekSheikh, line 292: In the derivation of cluster_potential above, please enhance the runtime_engine to be more complete. There are a lot of nulls at this point and as per the API docs, when it's null spark_version is used to determine the runtime_engine, so please add a rule to properly enhance the runtime_engine field when it is null.

After enhancing it in cluster_spec we should get great coverage but we'll still have a potential miss on the first run and/or if the value cannot be looked up from cluster_spec for some reason that's why I'm recommending to add this enhancement logic in both places.

src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
```diff
@@ -757,10 +757,11 @@ trait SilverTransforms extends SparkSessionWrapper {
...
           (unix_timestamp('Pipeline_SnapTS) * 1000).alias("timestamp"),
           'Pipeline_SnapTS.cast("date").alias("date"),
-          'creator_user_name.alias("createdBy")
+          'creator_user_name.alias("createdBy"),
+          'runtime_engine
```
GeekSheikh, line 764: There are a lot of nulls at this point and as per the API docs, when it's null spark_version is used to determine the runtime_engine, so please add a rule to properly enhance the runtime_engine field when it is null.

GeekSheikh (changes requested):

GeekSheikh (approved):
@souravbaner-da -- I had to make quite a few changes to even get the pipeline to run. Let's review testing methodologies again with the team asap. Slacked you the integration testing notebooks for PR_510, please review.

Issue #500, opened 2022-08-26
[Unity Catalog] Overwatch is failing with unity catalog


thetkoseek:
When using unity catalog, the schema cannot be defined a location. Because of this - it fails in the following area:
src/main/scala/com/databricks/labs/overwatch/pipeline/Initializer.scala

`
if (spark.catalog.databaseExists(dbName)) {
      val dbMeta = spark.sessionState.catalog.getDatabaseMetadata(dbName)
      val dbProperties = dbMeta.properties
      val existingDBLocation = dbMeta.locationUri.toString
      if (existingDBLocation != dbLocation) {
        switch = false
        throw new BadConfigException(s"The DB: $dbName exists " +
          s"at location $existingDBLocation which is different than the location entered in the config. Ensure " +
          s"the DBName is unique and the locations match. The location must be a fully qualified URI such as " +
          s"dbfs:/...")
      }
`

Would it be possible to bypass this check through a configuration parameter?

Note:
The whole pipeline from bronze,silver to gold works if the schema or database does not exist, as it does not do the check above. However when I run it again, it will do the location check and fail there as it expects the database to have a location defined.

GeekSheikh:
Hi @thetkoseek I believe you need follow the [external metastore](https://databrickslabs.github.io/overwatch/gettingstarted/advancedtopics/#using-external-metastores) rules when using UC. I have pinged two other colleagues that have gotten this to work on UC as well. They may be able to add additional context.

Note UC isn't fully supported yet but it does work with a few rules. We're working on getting these rules explicitly published in the docs.

You can follow #476 to see when full support is enabled.

GeekSheikh:
@thetkoseek -- The following appear to be the full list of requirements for Overwatch on UC. We will be getting docs out in the coming weeks and adding full support by end of year (sooner if possible) -- but in the meantime I wanted you and others with this issue to have an easy reference.

**UC REQUIREMENTS**
* DBR 11.1+
* Use full path (not mount point) as storage prefix
* Ensure the OW cluster has direct read/write access to the storage prefix
  * This should be set via `sc.hadoopConfiguration.set` and `spark.conf.set` in cluster config
* Ensure the UC has the appropriate permissions granted to all consumers

thetkoseek:
[...]

PR #511, opened 2022-09-06
Add check for if is unity catalog in the database location validation.

thetkoseek:
[https://github.com/databrickslabs/overwatch/issues/500](url)

There is no way to add a location property to a database in unity catalog and will always fail when validating whether the existing database location equals the specified location.

This PR will bypass that test if you add a `IS_UNITY_CATALOG ` property to the UC database


GeekSheikh:
Thank you for this. We have an entire sprint coming up to add full support and docs on UC enablement. For now, we have several customers using the default code base atop UC so long as they follow the rules discussed in #500 . If you cannot get that to work, please reopen #500 and we can discuss it further. My team can also give you a call to review your setup and ensure we have all the requirements we need for the upcoming sprint. 

Rejecting this for now as we're in process of building this and several other requirements for UC into the core.

PR #523, opened 2022-09-26
accumulator implemented

Sriram-databricks:
accumulator implemented for multithreaded cluster events api


GeekSheikh (approved):
gonna go ahead and merge this but if time ever permits it'd be awesome to get rid of those crazy sleeps. :P

PR #524, opened 2022-09-26
Dbsql cost 463

souravbaner-da:
Changed CLSF and deriveSKU for Adding cost for DBSQL


GeekSheikh:
cannot implement this until DB starts publishing warehouse events.

PR #526, opened 2022-09-26
api proxy implemented

Sriram-databricks:
api proxy implemented

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/ApiMeta.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/ApiMeta.scala
@@ -1,8 +1,11 @@
 package com.databricks.labs.overwatch
 
+import com.databricks.dbutils_v1.DBUtilsHolder.dbutils
+import com.databricks.labs.overwatch.utils.ApiEnv
 import com.fasterxml.jackson.databind.JsonNode
 import org.apache.log4j.{Level, Logger}
 import com.fasterxml.jackson.databind.JsonNode
+import scalaj.http.{Http, HttpRequest}
 
 /**
  * Configuration for each API.
@@ -17,7 +20,11 @@ trait ApiMeta {
   protected var _storeInTempLocation = false
   protected var _apiV = "api/2.0"
   protected var _isDerivePaginationLogic = false
+  protected var _apiEnv: ApiEnv = _
+  protected var _apiName: String = _
 
+  protected[overwatch] def apiName: String = _apiName
+  protected[overwatch] def apiEnv: ApiEnv = _apiEnv
   protected[overwatch] def paginationKey: String = _paginationKey
 
   protected[overwatch] def paginationToken: String = _paginationToken
@@ -34,6 +41,15 @@ trait ApiMeta {
 
   private[overwatch] def setApiV(value: String): this.type = {
     _apiV = value
+    this
+  }
+
+  private[overwatch] def setApiName(value: String): this.type = {
+    _apiName = value
+    this
+  }
+  private[overwatch] def setApiEnv(value: ApiEnv): this.type = {
+    _apiEnv = value
     this
   }
 
@@ -80,6 +96,31 @@ trait ApiMeta {
     true
   }
 
+  private[overwatch] def getBaseRequest(): HttpRequest = {
+    var request = Http(s"""${apiEnv.workspaceURL}/${apiV}/${apiName}""")
+      .copy(headers = httpHeaders)
+    if (apiEnv.proxyHost.isDefined && apiEnv.proxyPort.isDefined) {
+      request = request.proxy(apiEnv.proxyHost.get, apiEnv.proxyPort.get)
+      logger.log(Level.INFO, s"""Proxy has been set to IP: ${apiEnv.proxyHost.get}  PORT:${apiEnv.proxyPort.get}""")
+    }
+    if (apiEnv.proxyUserName.isDefined && apiEnv.proxyPasswordScope.isDefined && apiEnv.proxyPasswordKey.isDefined) {
+      val password = dbutils.secrets.get(scope = apiEnv.proxyPasswordScope.get, apiEnv.proxyPasswordKey.get)
+      request = request.proxyAuth(apiEnv.proxyUserName.get, password)
+      logger.log(Level.INFO, s"""Proxy UserName set to IP: ${apiEnv.proxyUserName.get}  scope:${apiEnv.proxyPasswordScope.get} key:${apiEnv.proxyPasswordKey.get}""")
+    }
+    request
+  }
+
+  /**
+   * Headers for the API call.
+   */
+  private def httpHeaders = Seq[(String, String)](
+    ("Content-Type", "application/json"),
+    ("Charset", "UTF-8"),
+    ("User-Agent", s"databricks-labs-overwatch-${apiEnv.packageVersion}"),
+    ("Authorization", s"Bearer ${apiEnv.rawToken}")
+  )
+
   override def toString: String = {
     s"""API Meta paginationKey: ${paginationKey}
        |paginationToken: ${paginationToken}
@@ -100,6 +141,7 @@ class ApiMetaFactory {
   private val logger: Logger = Logger.getLogger(this.getClass)
 
   def getApiClass(_apiName: String): ApiMeta = {
+
     val meta = _apiName match {
       case "jobs/list" => new JobListApi
       case "clusters/list" => new ClusterListApi
--- a/src/main/scala/com/databricks/labs/overwatch/pipeline/Initializer.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/pipeline/Initializer.scala
@@ -320,8 +320,8 @@ class Initializer(config: Config) extends SparkSessionWrapper {
         s"within the provided scope: ${tokenSecret.get.scope}. Please provide a scope and key " +
         s"available and accessible to this account.")
 
-      config.registerWorkspaceMeta(Some(TokenSecret(scopeName, keyCheck.head.key)))
-    } else config.registerWorkspaceMeta(None)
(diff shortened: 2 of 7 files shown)


src/main/scala/com/databricks/labs/overwatch/utils/Structures.scala
```diff
@@ -35,8 +35,37 @@ case class DatabricksContractPrices(
...
+                   proxyPort: Option[Int] = None,
+                   proxyUserName: Option[String] = None,
+                   proxyPasswordScope: Option[String] = None,
+                   proxyPasswordKey: Option[String] = None
+                 )
+
+
+case class ApiEnvConfig(
+                         successBatchSize: Int = 200,
+                         errorBatchSize: Int = 500,
```
GeekSheikh, line 59: I think these should be set == the defaults in `ApiEnv` and the defaults in Param Deserializer. Please unify the defaults

src/main/scala/com/databricks/labs/overwatch/pipeline/Initializer.scala
```diff
@@ -320,8 +320,8 @@ class Initializer(config: Config) extends SparkSessionWrapper {
         s"within the provided scope: ${tokenSecret.get.scope}. Please provide a scope and key " +
         s"available and accessible to this account.")
 
-      config.registerWorkspaceMeta(Some(TokenSecret(scopeName, keyCheck.head.key)))
-    } else config.registerWorkspaceMeta(None)
+      config.registerWorkspaceMeta(Some(TokenSecret(scopeName, keyCheck.head.key)),rawParams.apiEnvConfig)
+    } else config.registerWorkspaceMeta(None,None)
```
GeekSheikh, line 324: This part looks good, still looking for your validation section for the various configs.

src/main/scala/com/databricks/labs/overwatch/ApiMeta.scala
```diff
@@ -80,6 +96,31 @@ trait ApiMeta {
     true
   }
 
+  private[overwatch] def getBaseRequest(): HttpRequest = {
```
GeekSheikh, line 99: Nice! :) This is exactly what I was looking for. Good work.

GeekSheikh (changes requested):
very close, just a few minor outstanding items.

src/main/scala/com/databricks/labs/overwatch/pipeline/Initializer.scala
Sriram-databricks: done

src/main/scala/com/databricks/labs/overwatch/ApiMeta.scala
Sriram-databricks: thanks:)

src/main/scala/com/databricks/labs/overwatch/utils/Structures.scala
Sriram-databricks: done

Sriram-databricks (commented):
implemented the changes

src/main/scala/com/databricks/labs/overwatch/pipeline/Initializer.scala
```diff
@@ -237,6 +237,43 @@ class Initializer(config: Config) extends SparkSessionWrapper {
     config.setTempWorkingDir(workspaceTempWorkingDir)
   }
 
+
+  private def validateApiEnvConfig(apiEnvConfig: Option[ApiEnvConfig]) = {
```
GeekSheikh, line 241: please create valid ranges, not just ensuring it's >1 but it should be between some valid range. I mean the success batch size shouldn't be 200000000000 right, so just put in what you believe to be valid range

src/main/scala/com/databricks/labs/overwatch/pipeline/Initializer.scala
```diff
@@ -275,6 +312,7 @@ class Initializer(config: Config) extends SparkSessionWrapper {
       mapper.readValue[OverwatchParams](overwatchArgs)
     }
 
+    validateApiEnvConfig(rawParams.apiEnvConfig)
```
GeekSheikh, line 315: the validation should be done on the object after it's been created, not the rawParams. Please move this down a bit and utilize the constructed object for validation, not the raw values. This is important in case some setter has (or winds up with) logic that adjusts these values as the object[s] get built.

GeekSheikh (changes requested):

src/main/scala/com/databricks/labs/overwatch/pipeline/Initializer.scala
Sriram-databricks: sure, previously i thought of implementing it but i did not had any range to justify so removed it. Now i am adding it

src/main/scala/com/databricks/labs/overwatch/pipeline/Initializer.scala
Sriram-databricks: done

src/main/scala/com/databricks/labs/overwatch/utils/Config.scala
```diff
@@ -381,6 +382,43 @@ class Config() {
     }
   }
 
+  private def validateApiEnvConfig(apiEnvConfig: Option[ApiEnvConfig]) = {
+    if (apiEnvConfig.isDefined) {
+      val envConfig = apiEnvConfig.get
+      if (envConfig.threadPoolSize < 1 || envConfig.threadPoolSize > 20) {
+        throw new BadConfigException("ThreadPoolSize should be a valid number between 0 to 20")
+      }
+      if (envConfig.apiWaitingTime < 60000) { // 60000ms = 1 mint
```
GeekSheikh, line 391: should have a reasonable max -- say 15 mins? Does that seem right?

src/main/scala/com/databricks/labs/overwatch/utils/Config.scala
```diff
@@ -381,6 +382,43 @@ class Config() {
...
+      if (envConfig.successBatchSize < 1 || envConfig.successBatchSize > 1000) {
+        throw new BadConfigException("SuccessBatchSize should be between 1 to 1000")
+      }
+      if (envConfig.apiProxyConfig.isDefined) {
+        validateApiEnvProxy(envConfig.apiProxyConfig)
+      }
+    }
+  }
+
+  def validateApiEnvProxy(apiProxyConfig: Option[ApiProxyConfig]) = {
```
GeekSheikh, line 406: why did you move this to config? It doesn't belong in Config...right? I like how you switched it to validate the actual objects but no idea why you moved it here. :P let's keep this in Initializer and we'll get it cleaned up one of these days.

GeekSheikh (changes requested):

src/main/scala/com/databricks/labs/overwatch/utils/Config.scala
Sriram-databricks: The acutal object is present in Config.scala,but now Config.scala dont have access to Initialiser.scala object, we can not make call from Config.scala to Initialiser.scala methods. Thats why i moved the validation code to Config.scala.

src/main/scala/com/databricks/labs/overwatch/utils/Config.scala
Sriram-databricks: done

src/main/scala/com/databricks/labs/overwatch/utils/Config.scala
GeekSheikh: @Sriram-databricks but you have a getter and a setter right? If you don’t you should and then you can use that to get and validate from Initializer

Sriram-databricks: done

Sriram-databricks (commented):
implementation done

src/main/scala/com/databricks/labs/overwatch/utils/Config.scala
GeekSheikh: NOT FOR initializer! :P -- for `ApiEnvConfig` put a getter in Config and get it into the initializer after it's set and validate it in the initializer.

GeekSheikh (changes requested):
NOT FOR initializer! :P -- for ApiEnvConfig put a getter in Config and get it into the initializer after it's set and validate it in the initializer.

src/main/scala/com/databricks/labs/overwatch/pipeline/Initializer.scala
```diff
@@ -237,6 +237,32 @@ class Initializer(config: Config) extends SparkSessionWrapper {
...
+    if (apiEnv.apiWaitingTime < 60000 || apiEnv.apiWaitingTime > 900000) { // 60000ms = 1 mint,900000ms = 15mint
+      throw new BadConfigException("ApiWaiting time should be between 60000ms and 900000ms")
+    }
+    if (apiEnv.errorBatchSize < 1 || apiEnv.errorBatchSize > 1000) {
+      throw new BadConfigException("ErrorBatchSize should be between 1 to 1000")
+    }
+    if (apiEnv.successBatchSize < 1 || apiEnv.successBatchSize > 1000) {
+      throw new BadConfigException("SuccessBatchSize should be between 1 to 1000")
+    }
+    if (apiEnv.proxyHost.isDefined) {
```
GeekSheikh, line 253: `!isDefined` should always be `isEmpty` isDefined is for nulls, these are Options

Switch 
`isDefined` to `.nonEmpty` and
`!isDefined` to `.isEmpty` 

Like shown below. 

`apiEnv.proxyHost.isDefined` --> `apiEnv.proxyPort.nonEmpty`
`!apiEnv.proxyPort.isDefined` --> `apiEnv.proxyPort.isEmpty`

please do this everywhere an option type is referenced and you've used `isDefined`

GeekSheikh (changes requested):

GeekSheikh (approved):

PR #535, opened 2022-09-30
532 bug sql history

aman-db:
sqlHistory bug fix-
* providing filter (startTime and endTime) while using the sqlHistory API


GeekSheikh:
@Sriram-databricks -- Please help @aman-db breakdown and parallelize the APICall to SQLQuery History. 

In [getSqlQueryHistory](https://github.com/databrickslabs/overwatch/blob/a16b482cee95c3d83a7320c1f0a9a2316f924d58/src/main/scala/com/databricks/labs/overwatch/env/Workspace.scala#L132) the following `start_time_ms` and `end_time_ms` should never be great than 1 hour so if the from --> until time range is 7 days break it down into 1h API calls as Futures and send them in parallel to the `executeMultiThread`

GeekSheikh (approved):

PR #556, opened 2022-10-10
Sql history docs

aman-db:
Adding docs for sqlQueryHistory


.bsp/sbt.json
```diff
@@ -1 +0,0 @@
-{"name":"sbt","version":"1.4.9","bspVersion":"2.0.0-M5","languages":["scala"],"argv":["C:\\Program Files\\Java\\jdk1.8.0_271\\jre/bin/java","-Xms100m","-Xmx100m","-classpath","C:/Users/tomesd/AppData/Roaming/JetBrains/IntelliJIdea2021.1/plugins/Scala/launcher/sbt-launch.jar","xsbt.boot.Boot","-bsp"]}
```
GeekSheikh, line 1: This file needs to be removed from the commit tree.

GeekSheikh (changes requested):
looks good, just remove that .bsp folder from commit and re-request review and I'll recompile and merge. Thx

GeekSheikh (approved):
The TableSamples was in the wrong place, it must be in docs-compose/overwatch_docs/static/TableSamples -- EVERYTHING in the docs folder is generated by hugo so never put anything in there. I have made the change, but just FYI for future.

PR #569, opened 2022-10-18
[DQ] - Cluster Events - Merge Improvements

### LINKED ISSUE #485: [DQ] - cluster state detail silver - merge condition
The merge condition is keyed off of timestamp_start but a new column must be created and used as the key. 

When an OW run begins the state of a cluster is captured from where it left off until now. The current_state will not have an end_ts because it hasn't ended yet. To calculate costs through current OW run, the pipeline_snapshot time is used here. The next record on the subsequent OW run will be the end of the current state (from cluster_events_bronze) but it will be keyed in timestamp_state_start; thus the record doesn't get updated to reflect the proper end time resulting in the cluster appearing to be in two states at once until the pipeline_snapshot time of the previous run.

To correct this a new TS key field should be generated to key the merge and update the last record from the previous run. This solution needs to be reviewed and detailed out.

The reason this happens is because cluster_state_detail is driven by cluster_events_bronze. Each state from cluster_events_bronze ONLY has a `timestamp` the start/end are derived based off a lead/lag resulting in one record per key (i.e. cluster) getting extra time in state for each run.

image

### LINKED ISSUE #568: [DQ] - ClusterStateDetail - Missing Events
Relates to #485 

pipelineSnapTS was used as the end time instead of untilTimeTS. This should only result in a few missing events but can occasionally significantly skew costs.

GeekSheikh:
closes #568 
closes #485


souravbaner-da (approved):

GeekSheikh:
@souravbaner-da -- I believe all of these changes have already been merged into 07x -- could you take another look and just make sure there's nothing in this PR that didn't make it into 07x. If it looks like everything is there to you too -- we can go ahead and close this PR.

GeekSheikh:
all changes were pulled forward so this PR isn't necessary

PR #574, opened 2022-10-19
537 sql history parallelize

### LINKED ISSUE #537: [OPTIMIZE] - SQL Query History - Parallelize
Breakdown and parallelize the APICall to SQLQuery History.

In [getSqlQueryHistory](https://github.com/databrickslabs/overwatch/blob/3622f1cdba2142c95a22a72c848d50c0832dc9de/src/main/scala/com/databricks/labs/overwatch/env/Workspace.scala#L132) the following start_time_ms and end_time_ms should never be great than 1 hour so if the from --> until time range is 7 days break it down into 1h API calls as Futures and send them in parallel to the executeMultiThread

aman-db:
closes #537 
* Adding new function in workspace.scala for parallelize API call for sqlQueryHistory
* Replacing new function in Silver.scala for sqlQueryHistory

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/env/Workspace.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/env/Workspace.scala
@@ -167,17 +167,18 @@ class Workspace(config: Config) extends SparkSessionWrapper {
     implicit val ec: ExecutionContextExecutor = ExecutionContext.fromExecutor(Executors.newFixedThreadPool(config.apiEnv.threadPoolSize))
     val tmpSqlQueryHistorySuccessPath = s"${config.tempWorkingDir}/sqlqueryhistory_silver/${System.currentTimeMillis()}"
     val untilTimeMs = untilTime.asUnixTimeMilli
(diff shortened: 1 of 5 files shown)


src/main/scala/com/databricks/labs/overwatch/env/Workspace.scala
```diff
@@ -152,6 +158,116 @@ class Workspace(config: Config) extends SparkSessionWrapper {
...
+      future.onComplete {
+        case Success(_) =>
+          responseCounter = responseCounter + 1
+
+        case Failure(e) =>
+          if (e.isInstanceOf[ApiCallFailureV2]) {
+            synchronized {
+              apiErrorArray.add(e.getMessage)
+              if (apiErrorArray.size() >= config.apiEnv.errorBatchSize) {
+                PipelineFunctions.writeMicroBatchToTempLocation(tmpSqlQueryHistoryErrorPath, apiErrorArray.toString)
```
Sriram-databricks, line 226: What we are doing with the error data?

PR #615, opened 2022-11-09
ClusterEvents - Expanded Disk AFTER termination

souravbaner-da:
Change cluster state silver table to add the state Expanded Disk


GeekSheikh (approved):
LGTM

PR #616, opened 2022-11-09
adding sqlHistory Module in the dependency table

aman-db:
* adding sqlHistory module in dependency table


GeekSheikh (approved):
LGTM thx -- compiled.

Issue #505, opened 2022-09-01
Retry implementation for api 2.1 with 5** response code


Sriram-databricks:
Implement retry mechanism for api 2.1 for 5** response code

PR #626, opened 2022-11-21
APIv2 code change

### LINKED ISSUE #554: [BUG] - ApiCallV2 tempSuccessPath must be defaulted
When tempSuccessPath is not provided it must be defaulted, not throw a null pointer. Same with accumulator. Back to why I hate you setting null defaults :P 

```scala
  def apply(apiEnv: ApiEnv, apiName: String, queryJsonString: String) = {
    new ApiCallV2(apiEnv)
      .buildApi(apiName)
      .setQuery(queryJsonString)
  }
```

image

image

Sriram-databricks:
#505 closed
#554 closed


GeekSheikh (approved):
Nice work

Issue #567, opened 2022-10-18
[FEAT] Make the parallel API calls more generalize. 


aman-db:
Currently in order to make the parallel API calls for any module, we have to write the redundant code with very less changes. These parallel API calls can be generalized into a function/utils and can be used across the framework. 

* [ ] clusterEvents
* [ ] sqlQueryHistory
* [ ] runs failure out `jobs/runs/get-output` (coming soon)

GeekSheikh:
Adding jobs/runs/get-output api endpoint which is parallelized with key (similar to clusterEvents) -- this must be done before we can integrate this.

PR #635, opened 2022-11-28
537 Prallelize sqlQueryHistory API calls

### LINKED ISSUE #537: [OPTIMIZE] - SQL Query History - Parallelize

aman-db:
closes #537 
* Adding new function in workspace.scala for parallelize API call for sqlQueryHistory
* Replacing new function in Silver.scala for sqlQueryHistory

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/env/Workspace.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/env/Workspace.scala
@@ -2,13 +2,20 @@ package com.databricks.labs.overwatch.env
 
 import com.databricks.dbutils_v1.DBUtilsHolder.dbutils
 import com.databricks.labs.overwatch.ApiCallV2
+import com.databricks.labs.overwatch.pipeline.PipelineFunctions
 import com.databricks.labs.overwatch.utils._
 import org.apache.log4j.{Level, Logger}
 import org.apache.spark.sql.DataFrame
 import org.apache.spark.sql.functions._
 
+import java.util
+import java.util.Collections
+import java.util.concurrent.Executors
 import scala.collection.parallel.ForkJoinTaskSupport
+import scala.concurrent.{ExecutionContext, ExecutionContextExecutor, Future}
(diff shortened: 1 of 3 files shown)


src/main/scala/com/databricks/labs/overwatch/env/Workspace.scala
```diff
@@ -152,6 +159,132 @@ class Workspace(config: Config) extends SparkSessionWrapper {
...
+      try {
+        spark.read.json(tmpSqlQueryHistorySuccessPath)
+          .select(explode(col("res")).alias("res")).select(col("res" + ".*"))
+          .withColumn("organization_id", lit(config.organizationId))
+      } catch {
+        case e: Throwable =>
+          throw new Exception(e)
+      }
+    } else {
+      println("No Data is present in temporary Path")
```
Sriram-databricks, line 283: Can you print like no data found for sql/history/queries fromTimeMs - ToTimems

Sriram-databricks (changes requested):

src/main/scala/com/databricks/labs/overwatch/env/Workspace.scala
aman-db: done

Sriram-databricks (approved):

GeekSheikh (approved):
Approving this but we also prioritized #567 to 0702 to reduce the complexity of code like this. So after we generalize away most of this mess we can also simplify this code. For now, approving due to urgency. Good work.

PR #640, opened 2022-12-01
7004 release docs added

Sriram-databricks:
Public docs for 7004 release


GeekSheikh (approved):
Everything was good except you forgot to add the historical data fix notebooks to this PR. I went ahead and added them for us. Please verify I added the right one @Sriram-databricks

Issue #641, opened 2022-12-02
[UPDATE] - default hourly DBU costs - instanceDetails


GeekSheikh:
We've had multiple reports that several of the very common node types are showing incorrect dbu/hr costs lately. It seems Databricks has made substantial changes in the last month or so to hourly dbus by node type. 

We need to publish an update to the default node types hourly dbus. Furthermore, we need to add an FAQ on how to reset the instanceDetails table. This FAQ should provide a script that does the following
- drop table and delete data for instanceDetails (with big bold warning that customizations will be lost if this approach is taken)
  - If customer does not want to lose customizations they need to manually merge the new CSV with the existing Delta Table and only update hourly DBUs
- Rebuild CLSF and JRCP

The FAQ should also describe the reason we do not update the instanceDetails table automatically (customized value overwrites). Also note that we're working on finding / building a more automated process for this.

This ticket should result in two PRs, 1 for the updated CSVs and one for the docs update.

GeekSheikh:
Closed with #657

PR #658, opened 2022-12-14
Update FAQ.md

mohanbaabu1996:
Update FAQ.md

Issue #641


GeekSheikh:
Not doing this because I'm going to add it into the package in 0721 and make a function called Helpers.updateInstanceDetails() or something like that.

Issue #646, opened 2022-12-05
column cluster_type is missing from the view jobruncostpotentialfact


hollypinksmith:
cluster_type is missing from the view jobruncostpotentialfact, but is in the data underneath at jobruncostpotentialfact_gold

PR #673, opened 2023-01-03
adding cluster_type column in jrcp view

aman-db:
* adding column cluster_type in jrcp view
* change in StringExtTest


aman-db:
closes https://github.com/databrickslabs/overwatch/issues/646

PR #674, opened 2023-01-03
462 dbsql warehouse

### LINKED ISSUE #462: [FEAT] - Add Warehouses To Data Model
Let's get endpoints into the data model.

aman-db:
closes #462 
contains code for dbsql warehouse(endpoints)

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/pipeline/DbsqlTransforms.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/pipeline/DbsqlTransforms.scala
@@ -1,0 +1,130 @@
+package com.databricks.labs.overwatch.pipeline
+
+import com.databricks.labs.overwatch.pipeline.PipelineFunctions.fillForward
+import com.databricks.labs.overwatch.pipeline.TransformFunctions._
+import com.databricks.labs.overwatch.utils.SchemaTools.structFromJson
+import com.databricks.labs.overwatch.utils._
+import org.apache.log4j.{Level, Logger}
+import org.apache.spark.sql.expressions.{Window, WindowSpec}
+import org.apache.spark.sql.functions._
+import org.apache.spark.sql.types._
+import org.apache.spark.sql.{Column, DataFrame}
+
+object DbsqlTransforms extends SparkSessionWrapper {
+
+  import spark.implicits._
+  private val logger: Logger = Logger.getLogger(this.getClass)
+  val responseSuccessFilter: Column = $"response.statusCode" === 200
+
+  /**
+   * BEGIN DBSQL generic functions
+   */
+
+  private val auditBaseCols: Array[Column] = Array(
+    'timestamp, 'date, 'organization_id, 'serviceName, 'actionName,
+    $"userIdentity.email".alias("userEmail"), 'requestId, 'response)
+
+  def deriveWarehouseId(): Column = {
+    when(('actionName === "createEndpoint" || 'actionName === "createWarehouse")
+      && responseSuccessFilter,
+      get_json_object($"response.result", "$.id"))
+      .otherwise('id)
+  }
+
+  def deriveWarehouseBase(auditRawDF: DataFrame): DataFrame = {
+    val warehouse_id_gen_w = Window.partitionBy('organization_id, 'warehouse_name)
+      .orderBy('timestamp).rowsBetween(Window.currentRow, 1000)
+    val warehouse_name_gen_w = Window.partitionBy('organization_id, 'warehouse_id)
+      .orderBy('timestamp).rowsBetween(Window.currentRow, 1000)
+
+    val warehouseSummaryCols = auditBaseCols ++ Array[Column](
+      deriveWarehouseId.alias("warehouse_id"),
+      'name.alias("warehouse_name"),
+      'cluster_size,
+      'min_num_clusters.cast("long"),
+      'max_num_clusters.cast("long"),
+      'auto_stop_mins.cast("long"),
+      'spot_instance_policy,
+      'enable_photon.cast("boolean"),
+      'channel,
+      'tags,
+      'enable_serverless_compute.cast("boolean"),
+      'warehouse_type
+    )
+
+    val warehouseRaw = auditRawDF
+      .filter('serviceName === "databrickssql")
+      .selectExpr("*", "requestParams.*").drop("requestParams", "Overwatch_RunID")
+      .select(warehouseSummaryCols: _*)
+      .withColumn("warehouse_id", PipelineFunctions.fillForward("warehouse_id", warehouse_id_gen_w))
+      .withColumn("warehouse_name", PipelineFunctions.fillForward("warehouse_name", warehouse_name_gen_w))
+
+    val warehouseWithStructs = warehouseRaw
+      .withColumn("channel", SchemaTools.structFromJson(spark, warehouseRaw, "channel"))
+      .withColumn("tags", SchemaTools.structFromJson(spark, warehouseRaw, "tags"))
+      .scrubSchema
+
+    warehouseWithStructs
+      .withColumn("tags", SchemaTools.structToMap(warehouseWithStructs, "tags"))
+  }
+
+  /**
+   * warehouseBaseFilled - if first run, baseline warehouse spec for existing warehouses that haven't been edited since
+   * commencement of audit logs. Allows for joins directly to gold warehouse work even if they haven't yet been edited.
+   * Several of the fields are unavailable through this method but many are and they are very valuable when
+   * present in gold
+   */
+  def deriveWarehouseBaseFilled(isFirstRun: Boolean, bronzeWarehouseSnapUntilCurrent: DataFrame)
+                                (warehouseBaseWMetaDF: DataFrame): DataFrame = {
+    if (isFirstRun) {
+      val firstRunMsg = "Silver_WarehouseSpec -- First run detected, will impute warehouse state from bronze to derive " +
+        "current initial state for all existing warehouses."
+      logger.log(Level.INFO, firstRunMsg)
+      println(firstRunMsg)
+      val missingWarehouseIds = bronzeWarehouseSnapUntilCurrent.select('organization_id,
+        'warehouse_id).distinct
+        .join(
+          warehouseBaseWMetaDF
+            .select('organization_id, 'warehouse_id).distinct,
+          Seq("organization_id", "warehouse_id"), "anti"
+        )
+      val latestWarehouseSnapW = Window.partitionBy('organization_id, 'warehouse_id).orderBy('Pipeline_SnapTS.desc)
+
+      val missingWareHouseBaseFromSnap = bronzeWarehouseSnapUntilCurrent
+        .join(missingWarehouseIds, Seq("organization_id", "warehouse_id"))
+        .withColumn("rnk", rank().over(latestWarehouseSnapW))
+        .filter('rnk === 1).drop("rnk")
+        .select(
+          'organization_id,
+          'warehouse_id,
+          lit("warehouses").alias("serviceName"),
+          lit("snapImpute").alias("actionName"),
+          'name.alias("warehouse_name"),
+          'state.alias("warehouse_state"),
+          'size,
+          'cluster_size,
+          'min_num_clusters,
+          'max_num_clusters,
+          'auto_stop_mins,
+          'auto_resume,
+          'creator_id,
+          'spot_instance_policy,
+          'enable_photon,
+          'channel,
+          'tags,
+          'enable_serverless_compute,
+          'warehouse_type,
+          'num_clusters,
+          'num_active_sessions,
+          'jdbc_url,
+          'odbc_params,
+          (unix_timestamp('Pipeline_SnapTS) * 1000).alias("timestamp"),
+          'Pipeline_SnapTS.cast("date").alias("date"),
+          'creator_name.alias("createdBy")
+        )
+      unionWithMissingAsNull(warehouseBaseWMetaDF, missingWareHouseBaseFromSnap)
+    } else warehouseBaseWMetaDF
+  }
+
+
+}
--- a/src/main/scala/com/databricks/labs/overwatch/pipeline/DbsqlTransforms.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/pipeline/DbsqlTransforms.scala
@@ -20,9 +20,9 @@ object DbsqlTransforms extends SparkSessionWrapper {
    * BEGIN DBSQL generic functions
    */
 
-  private val auditBaseCols: Array[Column] = Array(
-    'timestamp, 'date, 'organization_id, 'serviceName, 'actionName,
-    $"userIdentity.email".alias("userEmail"), 'requestId, 'response)
+//  private val auditBaseCols: Array[Column] = Array(
+//    'timestamp, 'date, 'organization_id, 'serviceName, 'actionName,
+//    $"userIdentity.email".alias("userEmail"), 'requestId, 'response)
 
   def deriveWarehouseId(): Column = {
     when(('actionName === "createEndpoint" || 'actionName === "createWarehouse")
@@ -31,7 +31,7 @@ object DbsqlTransforms extends SparkSessionWrapper {
       .otherwise('id)
   }
 
-  def deriveWarehouseBase(auditRawDF: DataFrame): DataFrame = {
+  def deriveWarehouseBase(auditRawDF: DataFrame, auditBaseCols: Array[Column]): DataFrame = {
     val warehouse_id_gen_w = Window.partitionBy('organization_id, 'warehouse_name)
       .orderBy('timestamp).rowsBetween(Window.currentRow, 1000)
     val warehouse_name_gen_w = Window.partitionBy('organization_id, 'warehouse_id)
@@ -41,14 +41,14 @@ object DbsqlTransforms extends SparkSessionWrapper {
       deriveWarehouseId.alias("warehouse_id"),
       'name.alias("warehouse_name"),
       'cluster_size,
-      'min_num_clusters.cast("long"),
-      'max_num_clusters.cast("long"),
-      'auto_stop_mins.cast("long"),
+      'min_num_clusters,
+      'max_num_clusters,
+      'auto_stop_mins,
       'spot_instance_policy,
-      'enable_photon.cast("boolean"),
+      'enable_photon,
       'channel,
       'tags,
-      'enable_serverless_compute.cast("boolean"),
+      'enable_serverless_compute,
       'warehouse_type
     )
 
@@ -121,7 +121,7 @@ object DbsqlTransforms extends SparkSessionWrapper {
           (unix_timestamp('Pipeline_SnapTS) * 1000).alias("timestamp"),
           'Pipeline_SnapTS.cast("date").alias("date"),
           'creator_name.alias("createdBy")
-        )
+        ).unionByName(warehouseBaseWMetaDF, allowMissingColumns = true)
       unionWithMissingAsNull(warehouseBaseWMetaDF, missingWareHouseBaseFromSnap)
     } else warehouseBaseWMetaDF
   }
--- a/src/main/scala/com/databricks/labs/overwatch/pipeline/DbsqlTransforms.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/pipeline/DbsqlTransforms.scala
@@ -20,43 +20,39 @@ object DbsqlTransforms extends SparkSessionWrapper {
    * BEGIN DBSQL generic functions
    */
 
-//  private val auditBaseCols: Array[Column] = Array(
-//    'timestamp, 'date, 'organization_id, 'serviceName, 'actionName,
-//    $"userIdentity.email".alias("userEmail"), 'requestId, 'response)
-
   def deriveWarehouseId(): Column = {
     when(('actionName === "createEndpoint" || 'actionName === "createWarehouse"),
       get_json_object($"response.result", "$.id"))
       .otherwise('id)
   }
 
-  def deriveWarehouseBase(auditRawDF: DataFrame, auditBaseCols: Array[Column]): DataFrame = {
-    val warehouse_id_gen_w = Window.partitionBy('organization_id, 'warehouse_name)
-      .orderBy('timestamp).rowsBetween(Window.currentRow, 1000)
+  def deriveWarehouseBase()(auditRawDF: DataFrame): DataFrame = {
     val warehouse_name_gen_w = Window.partitionBy('organization_id, 'warehouse_id)
-      .orderBy('timestamp).rowsBetween(Window.currentRow, 1000)
-
-    val warehouseSummaryCols = auditBaseCols ++ Array[Column](
-      deriveWarehouseId.alias("warehouse_id"),
-      'name.alias("warehouse_name"),
-      'cluster_size,
-      'min_num_clusters,
-      'max_num_clusters,
-      'auto_stop_mins,
-      'spot_instance_policy,
-      'enable_photon,
-      get_json_object('channel, "$.name").alias("channel"),
-      'tags,
-      'enable_serverless_compute,
-      'warehouse_type
-    )
+      .orderBy('timestamp).rowsBetween(Window.unboundedPreceding, 1000)
 
     val warehouseRaw = auditRawDF
-      .filter('serviceName === "databrickssql")
-      .selectExpr("*", "requestParams.*").drop("requestParams", "Overwatch_RunID")
-      .select(warehouseSummaryCols: _*)
-      .withColumn("warehouse_id", PipelineFunctions.fillForward("warehouse_id", warehouse_id_gen_w))
-      .withColumn("warehouse_name", PipelineFunctions.fillForward("warehouse_name", warehouse_name_gen_w))
+      .select(
+        'timestamp,
+        'date,
+        'serviceName,
+        'actionName,
+        'userEmail,
+        'requestId,
+        'response,
+        'warehouse_id,
+        PipelineFunctions.fillForward("warehouse_name",warehouse_name_gen_w),
+        PipelineFunctions.fillForward("cluster_size",warehouse_name_gen_w),
+        PipelineFunctions.fillForward("min_num_clusters",warehouse_name_gen_w),
+        PipelineFunctions.fillForward("max_num_clusters",warehouse_name_gen_w),
+        PipelineFunctions.fillForward("auto_stop_mins",warehouse_name_gen_w),
+        PipelineFunctions.fillForward("spot_instance_policy",warehouse_name_gen_w),
+        PipelineFunctions.fillForward("enable_photon",warehouse_name_gen_w),
+        PipelineFunctions.fillForward("channel",warehouse_name_gen_w),
+        PipelineFunctions.fillForward("tags",warehouse_name_gen_w),
+        PipelineFunctions.fillForward("enable_serverless_compute",warehouse_name_gen_w),
+        PipelineFunctions.fillForward("warehouse_type",warehouse_name_gen_w)
+      )
+
 
     val warehouseWithStructs = warehouseRaw
       .withColumn("tags", SchemaTools.structFromJson(spark, warehouseRaw, "tags"))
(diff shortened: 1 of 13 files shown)


src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
```diff
@@ -1314,4 +1314,119 @@ trait SilverTransforms extends SparkSessionWrapper {
...
+    val warehouse_name_gen_w = Window.partitionBy('organization_id, 'warehouse_id)
+      .orderBy('timestamp).rowsBetween(Window.currentRow, 1000)
+    val warehouse_id_gen = first('warehouse_id, true).over(warehouse_id_gen_w)
+    val warehouse_name_gen = first('warehouse_name, true).over(warehouse_name_gen_w)
+
+    val warehouseSummaryCols = auditBaseCols ++ Array[Column](
+      when('actionName === "createEndpoint" || 'actionName === "createWarehouse",
+        get_json_object($"response.result", "$.id"))
+        .when(('actionName =!= "createEndpoint" && 'id.isNull)
+          || ('actionName =!= "createWarehouse" && 'id.isNull), 'warehouseId)
```
GeekSheikh, line 1329: the most exclusive must come first in your case statement or this case will never be hit since the one above it will match first.

GeekSheikh: please move this logic to a func like

`private def deriveWarehouseId(df: DataFrame) = when...when...otherwise...alias("warehouse_id")`

and then use this func as one of your selects. This will allow us to add unit tests in the future.

src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
```diff
@@ -1314,4 +1314,119 @@ trait SilverTransforms extends SparkSessionWrapper {
 
   }
 
+  private def warehouseBase(auditRawDF: DataFrame): DataFrame = {
+    val warehouse_id_gen_w = Window.partitionBy('organization_id, 'warehouse_name)
+      .orderBy('timestamp).rowsBetween(Window.currentRow, 1000)
+    val warehouse_name_gen_w = Window.partitionBy('organization_id, 'warehouse_id)
+      .orderBy('timestamp).rowsBetween(Window.currentRow, 1000)
+    val warehouse_id_gen = first('warehouse_id, true).over(warehouse_id_gen_w)
+    val warehouse_name_gen = first('warehouse_name, true).over(warehouse_name_gen_w)
```
GeekSheikh, line 1323: This logic appears unnecessary and incorrect. Warehouse_id is a key -- should we really be filling like this? Are there really actions in scope that result in no warehouse id?

Warehouse_name -- I see that you're trying to fill forward -- that makes sense -- but please use the appropriate fillForward commands -- this is not accurate because your getting the first non_null warehouse name and it's not even ordered. Additionally, this likely won't help you at all since you're not pulling in lagging history so you likely won't have any previous data in many records. Please reference the other fillForward examples and consider pulling in an appropriate amount of historical if you intend to fill. We can discuss further if you like.

src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
```diff
@@ -1314,4 +1314,119 @@ trait SilverTransforms extends SparkSessionWrapper {
...
+
+    val warehouseSummaryCols = auditBaseCols ++ Array[Column](
```
GeekSheikh, line 1325: please make the timestamp field unixTimeMS as it is in other silver tables and make timestamp == an actual timestamp type for humans. The key for this and the gold target should be switched to the unixTimeMS field.

src/main/scala/com/databricks/labs/overwatch/pipeline/PipelineTargets.scala
```diff
@@ -356,6 +366,14 @@ abstract class PipelineTargets(config: Config) {
       partitionBy = Seq("organization_id")
     )
 
+    lazy private[overwatch] val warehousesSpecTarget: PipelineTable = PipelineTable(
+      name = "warehouse_spec_silver",
+      _keys = Array("timestamp", "warehouse_id"),
```
GeekSheikh, line 371: switch key from timestamp to unixTimeMS

src/test/scala/com/databricks/labs/overwatch/utils/StringExtTest.scala
```diff
@@ -27,11 +29,11 @@ class StringExtTest extends AnyFunSpec {
...
-      assert(!"""Business %Name""".containsNoSpecialChars)
+      assert("""Business %Name""".containsNoSpecialChars)
     }
 
     it("String with white $ characters") {
-      assert(!"""Business $Name""".containsNoSpecialChars)
+      assert("""Business $Name""".containsNoSpecialChars)
     }
   }
 }
```
GeekSheikh, line 39: did you mean to leave these changes or are these left-overs from the previous PR? Please drop these changes if they were unintentional.

src/main/scala/com/databricks/labs/overwatch/pipeline/Silver.scala
```diff
@@ -368,6 +372,21 @@ class Silver(_workspace: Workspace, _database: Database, _config: Config)
       )
   }
 
+  lazy private[overwatch] val warehouseSpecModule = Module(2021, "Silver_WarehouseSpec", this, Array(1004))
+  lazy private val appendWarehouseSpecProcess: () => ETLDefinition = {
+    () =>
+    ETLDefinition(
+      BronzeTargets.auditLogsTarget.asIncrementalDF(warehouseSpecModule, BronzeTargets.auditLogsTarget.incrementalColumns),
+      Seq(
+        buildWarehouseSpec(
```
GeekSheikh, line 381: I think you're going to need to send in full silver warehouse target itself so you can fill forward appropriately.

src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
```diff
@@ -1314,4 +1314,119 @@ trait SilverTransforms extends SparkSessionWrapper {
...
+
+    warehouseWithStructs
+        .withColumn("tags", SchemaTools.structToMap(warehouseWithStructs, "tags"))
+  }
+
+  protected def buildWarehouseSpec(
+                                  bronze_warehouse_snap: PipelineTable,
+                                  isFirstRun: Boolean,
+                                  untilTime: TimeTypes
+                                )(df: DataFrame): DataFrame = {
```
GeekSheikh, line 1364: do me a favor and refactor this like the workflowsTransforms. I want a dbsqlTransforms.scala object that contains all the transforms. This should have been the pattern to begin with.

This will allow you to do something like:

```scala
deriveWarehouseBase(df)
  .transform(initializeFirstRuns(df1, df2...))
  .transform(...)...
```

GeekSheikh, line 1364: do me a favor and refactor this like the workflowsTransforms. I want a dbsqlTransforms.scala object that contains all the transforms. This should have been the pattern to begin with.

This will allow you to do something like:

```scala
deriveWarehouseBase(df)
  .transform(initializeFirstRuns(df1, df2...))
  .transform(...)...
```

src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
```diff
@@ -1314,4 +1314,119 @@ trait SilverTransforms extends SparkSessionWrapper {
...
+            .select('organization_id, 'warehouse_id).distinct,
+          Seq("organization_id", "warehouse_id"), "anti"
+        )
+      val latestWarehouseSnapW = Window.partitionBy('organization_id, 'warehouse_id).orderBy('Pipeline_SnapTS.desc)
+
+      val missingWareHouseBaseFromSnap = bronzeWarehouseSnapUntilCurrent
+        .join(missingWarehouseIds, Seq("organization_id", "warehouse_id"))
+        .withColumn("rnk", rank().over(latestWarehouseSnapW))
+        .filter('rnk === 1).drop("rnk")
+        .select(
```
GeekSheikh, line 1397: you're force-casting several of these types from the audit logs -- be sure the types match on all the fields. I'm guessing you've already verified this via the WarehouseSnap Schema, just want to make sure.

GeekSheikh (changes requested):
good job on hitting most of the points on a full new pipeline. I left some comments and am going to need some more due diligence from you on the proof. Will discuss further in slack.

src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
aman-db: As discussed , we will keep timestamp field in Silver table. As other Silver tables has timestamp  column too.

src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
aman-db: Yes we have found null warehouse ids. Thats why we are filling it forward

src/main/scala/com/databricks/labs/overwatch/pipeline/Schema.scala
```diff
@@ -762,6 +787,8 @@ object Schema extends SparkSessionWrapper {
     2017 -> auditMasterSchema,
     // Notebook Summary
     2018 -> auditMasterSchema,
+    // Warehouse Spec
+    2021 -> auditMasterSchema,
```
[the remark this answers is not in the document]
aman-db: done

src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
aman-db: done. created a dbsqlTransform.scala

src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
aman-db: done

aman-db: done

src/test/scala/com/databricks/labs/overwatch/utils/StringExtTest.scala
aman-db: done

souravbaner-da (changes requested):
@aman-db  In the PR Proof please run one resume pipeline for OW so that we can check whether the module is working fine in 2nd and consecutive runs.

aman-db:
@souravbaner-da - done. Ran the deployment twice. Please check now

src/main/scala/com/databricks/labs/overwatch/pipeline/DbsqlTransforms.scala
```diff
@@ -0,0 +1,130 @@
...
+      .otherwise('id)
+  }
+
+  def deriveWarehouseBase(auditRawDF: DataFrame): DataFrame = {
+    val warehouse_id_gen_w = Window.partitionBy('organization_id, 'warehouse_name)
+      .orderBy('timestamp).rowsBetween(Window.currentRow, 1000)
+    val warehouse_name_gen_w = Window.partitionBy('organization_id, 'warehouse_id)
+      .orderBy('timestamp).rowsBetween(Window.currentRow, 1000)
+
+    val warehouseSummaryCols = auditBaseCols ++ Array[Column](
```
GeekSheikh, line 40: don't cast here -- all of these fields should be in Schema.scala and typed correctly already. If they're landed and typed correctly in audit_log_bronze, you don't need to cast them here.

src/main/scala/com/databricks/labs/overwatch/pipeline/DbsqlTransforms.scala
```diff
@@ -0,0 +1,130 @@
...
+      'enable_serverless_compute.cast("boolean"),
+      'warehouse_type
+    )
+
+    val warehouseRaw = auditRawDF
+      .filter('serviceName === "databrickssql")
+      .selectExpr("*", "requestParams.*").drop("requestParams", "Overwatch_RunID")
+      .select(warehouseSummaryCols: _*)
+      .withColumn("warehouse_id", PipelineFunctions.fillForward("warehouse_id", warehouse_id_gen_w))
+      .withColumn("warehouse_name", PipelineFunctions.fillForward("warehouse_name", warehouse_name_gen_w))
```
GeekSheikh, line 60: no. First, warehouse_id should never be null; thus you don't want to try and derive id from name. Warehouse_name may be necessary to fillForward but only if there are nulls for some of the actionNames in scope. Also, if name is null; there are likely several other fields that are null as well and for those you'll need to fill forward but you shouldn't do this with a withColumn each time as that creates a very nasty spark plan. Each of these fillForward functions should be part of the selects. If any are still unfillable you can use the warehouseSnap as a secondary fill. See the `poolSelects` in SilverTransforms.scala as an example.

src/main/scala/com/databricks/labs/overwatch/pipeline/DbsqlTransforms.scala
```diff
@@ -0,0 +1,130 @@
...
+
+    val warehouseWithStructs = warehouseRaw
+      .withColumn("channel", SchemaTools.structFromJson(spark, warehouseRaw, "channel"))
```
GeekSheikh, line 63: I recommended you just grab the json object of name. Looks like that's the only value present so suggest you use that.

src/main/scala/com/databricks/labs/overwatch/pipeline/DbsqlTransforms.scala
```diff
@@ -0,0 +1,130 @@
...
+    val warehouseSummaryCols = auditBaseCols ++ Array[Column](
+      deriveWarehouseId.alias("warehouse_id"),
+      'name.alias("warehouse_name"),
+      'cluster_size,
+      'min_num_clusters.cast("long"),
+      'max_num_clusters.cast("long"),
+      'auto_stop_mins.cast("long"),
+      'spot_instance_policy,
+      'enable_photon.cast("boolean"),
+      'channel,
```
GeekSheikh, line 49: suggest `get_json_object('channel, "$.name")` as that seems to be the only populated value at this time.

src/main/scala/com/databricks/labs/overwatch/pipeline/DbsqlTransforms.scala
```diff
@@ -0,0 +1,130 @@
...
+   * BEGIN DBSQL generic functions
+   */
+
+  private val auditBaseCols: Array[Column] = Array(
+    'timestamp, 'date, 'organization_id, 'serviceName, 'actionName,
+    $"userIdentity.email".alias("userEmail"), 'requestId, 'response)
+
+  def deriveWarehouseId(): Column = {
+    when(('actionName === "createEndpoint" || 'actionName === "createWarehouse")
+      && responseSuccessFilter,
```
GeekSheikh, line 29: should the response filter go here? I think it should go on the first reference of auditRawDF. Seems like the logic is wrong to put the responseSuccessFulter here.

src/main/scala/com/databricks/labs/overwatch/pipeline/DbsqlTransforms.scala
```diff
@@ -0,0 +1,130 @@
...
+
+  import spark.implicits._
+  private val logger: Logger = Logger.getLogger(this.getClass)
+  val responseSuccessFilter: Column = $"response.statusCode" === 200
+
+  /**
+   * BEGIN DBSQL generic functions
+   */
+
+  private val auditBaseCols: Array[Column] = Array(
```
GeekSheikh, line 23: Shouldn't this be passed into the func and not defined as a class val here? It's duplicate code.

src/main/scala/com/databricks/labs/overwatch/pipeline/DbsqlTransforms.scala
```diff
@@ -0,0 +1,130 @@
...
+import org.apache.spark.sql.expressions.{Window, WindowSpec}
+import org.apache.spark.sql.functions._
+import org.apache.spark.sql.types._
+import org.apache.spark.sql.{Column, DataFrame}
+
+object DbsqlTransforms extends SparkSessionWrapper {
+
+  import spark.implicits._
+  private val logger: Logger = Logger.getLogger(this.getClass)
+  val responseSuccessFilter: Column = $"response.statusCode" === 200
```
GeekSheikh, line 17: Shouldn't this be passed into the func and not defined as a class val here? It's duplicate code.

src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
```diff
@@ -1353,4 +1354,63 @@ trait SilverTransforms extends SparkSessionWrapper {
...
+        .withColumn("tags", SchemaTools.structToMap(warehouseWithStructs, "tags"))
+  }
+
+  protected def buildWarehouseSpec(
+                                  bronze_warehouse_snap: PipelineTable,
+                                  isFirstRun: Boolean,
+                                  untilTime: TimeTypes
+                                )(df: DataFrame): DataFrame = {
+    val bronzeWarehouseSnapUntilCurrent = bronze_warehouse_snap.asDF
+      .filter('Pipeline_SnapTS <= untilTime.asColumnTS)
```
GeekSheikh, line 1407: this logic is not correct. You need the latest snap for each not all snaps prior to today. See `bronzeClusterSnapLatest` in SilverTransforms.scala

src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
```diff
@@ -1353,4 +1354,63 @@ trait SilverTransforms extends SparkSessionWrapper {
...
+
+    val filteredDf = df
+          .filter('actionName.isin("createEndpoint", "editEndpoint", "createWarehouse",
+            "editWarehouse", "deleteEndpoint", "deleteWarehouse"))
```
GeekSheikh, line 1411: I think this is where the responseSuccessFilter should be probably.

src/main/scala/com/databricks/labs/overwatch/pipeline/Gold.scala
```diff
@@ -66,7 +67,8 @@ class Gold(_workspace: Workspace, _database: Database, _config: Config)
         )
       }
       case OverwatchScope.dbsql => {
-        Array(sqlQueryHistoryModule)
+        Array(sqlQueryHistoryModule,
+          warehouseModule)
```
GeekSheikh, line 71: formatting -- add new line like sparkEvents

src/main/scala/com/databricks/labs/overwatch/pipeline/Schema.scala
```diff
@@ -208,6 +208,15 @@ object Schema extends SparkSessionWrapper {
...
         StructField("resourceId", LongType, nullable = true),
+        StructField("cluster_size", StringType, nullable = true),
+        StructField("min_num_clusters", LongType, nullable = true),
+        StructField("max_num_clusters", LongType, nullable = true),
+        StructField("auto_stop_mins", LongType, nullable = true),
+        StructField("spot_instance_policy", StringType, nullable = true),
+        StructField("enable_photon", BooleanType, nullable = true),
+        StructField("channel", StringType, nullable = true),
+        StructField("enable_serverless_compute", BooleanType, nullable = true),
+        StructField("warehouse_type", StringType, nullable = true),
```
GeekSheikh, line 219: good job -- I think you got all the fields you're using. Triple check to make sure they're all here AND that none of them are duplicated.

src/main/scala/com/databricks/labs/overwatch/pipeline/Schema.scala
```diff
@@ -725,6 +734,31 @@ object Schema extends SparkSessionWrapper {
     StructField("workspace_name", StringType, nullable = true)
   ))
 
+  val warehouseSnapMinimumSchema: StructType = StructType(Seq(
+    StructField("id", StringType, nullable = true),
```
GeekSheikh, line 738: this SnapMinSchema is the schema AT REST in the target. In the snap land in bronze you renamed `id` to `warehouse_id` so this should be warehouse_id and it should not be nullable since it's a key

GeekSheikh (changes requested):
close but needs another look.

GeekSheikh:
@aman-db -- don't forget to add your new module to `getTargetTableNameByModule` in PipelineFunctions.scala

src/main/scala/com/databricks/labs/overwatch/pipeline/DbsqlTransforms.scala
aman-db: done

src/main/scala/com/databricks/labs/overwatch/pipeline/DbsqlTransforms.scala
aman-db: done

src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
```diff
@@ -1353,4 +1354,63 @@ trait SilverTransforms extends SparkSessionWrapper {
 
   }
 
+  // Below function is deprecated and can be removed
+  private def warehouseBase(auditRawDF: DataFrame): DataFrame = {
```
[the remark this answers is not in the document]
aman-db: done

src/main/scala/com/databricks/labs/overwatch/pipeline/DbsqlTransforms.scala
```diff
@@ -0,0 +1,130 @@
...
+          'warehouse_type,
+          'num_clusters,
+          'num_active_sessions,
+          'jdbc_url,
+          'odbc_params,
+          (unix_timestamp('Pipeline_SnapTS) * 1000).alias("timestamp"),
+          'Pipeline_SnapTS.cast("date").alias("date"),
+          'creator_name.alias("createdBy")
+        )
+      unionWithMissingAsNull(warehouseBaseWMetaDF, missingWareHouseBaseFromSnap)
```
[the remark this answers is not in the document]
aman-db: done

src/main/scala/com/databricks/labs/overwatch/pipeline/Gold.scala
aman-db: done

src/main/scala/com/databricks/labs/overwatch/pipeline/Schema.scala
aman-db: done

src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
aman-db: done

src/main/scala/com/databricks/labs/overwatch/pipeline/DbsqlTransforms.scala
aman-db: done. Removed this filter from here

src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
aman-db: done. Added the filter

src/main/scala/com/databricks/labs/overwatch/pipeline/DbsqlTransforms.scala
aman-db: done. Removed from here and added at auditRawDf

src/main/scala/com/databricks/labs/overwatch/pipeline/DbsqlTransforms.scala
aman-db: done. Removed it from here and added in the warehouseSummaryCols.

src/main/scala/com/databricks/labs/overwatch/pipeline/DbsqlTransforms.scala
aman-db: done. Also passing full silver_spec_warehouse to fill forward appropriately.

src/main/scala/com/databricks/labs/overwatch/pipeline/Schema.scala
```diff
@@ -725,6 +734,31 @@ object Schema extends SparkSessionWrapper {
     StructField("workspace_name", StringType, nullable = true)
   ))
 
+  val warehouseSnapMinimumSchema: StructType = StructType(Seq(
```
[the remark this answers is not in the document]
aman-db: done

souravbaner-da (approved):

src/main/scala/com/databricks/labs/overwatch/pipeline/Schema.scala
```diff
@@ -725,6 +734,31 @@ object Schema extends SparkSessionWrapper {
...
+    StructField("spot_instance_policy", StringType, nullable = true),
+    StructField("enable_photon", BooleanType, nullable = true),
+    StructField("enable_serverless_compute", BooleanType, nullable = true),
+    StructField("warehouse_type", StringType, nullable = true),
+    StructField("num_clusters", LongType, nullable = true),
+    StructField("num_active_sessions", LongType, nullable = true),
+    StructField("state", StringType, nullable = true),
+    StructField("organization_id", StringType, nullable = false),
+    StructField("Pipeline_SnapTS", TimestampType, nullable = true),
+    StructField("Overwatch_RunID", StringType, nullable = true),
```
gueniai, line 758: Overwatch_RunID is part of your target key, I don't think it should be nullable. I also think Pipeline_SnapTS shouldn't be nullable either

src/main/scala/com/databricks/labs/overwatch/pipeline/Silver.scala
```diff
@@ -371,6 +375,21 @@ class Silver(_workspace: Workspace, _database: Database, _config: Config)
       )
   }
 
+  lazy private[overwatch] val warehouseSpecModule = Module(2021, "Silver_WarehouseSpec", this, Array(1004))
```
gueniai, line 378: Also dependant on 1013

src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
```diff
@@ -1353,4 +1354,21 @@ trait SilverTransforms extends SparkSessionWrapper {
...
+  protected def buildWarehouseSpec(
+                                  bronze_warehouse_snap: PipelineTable,
+                                  isFirstRun: Boolean,
+                                  silver_warehouse_spec: PipelineTable
+                                )(df: DataFrame): DataFrame = {
+    val lastWarehouseSnapW = Window.partitionBy('organization_id, 'warehouse_id)
+      .orderBy('Pipeline_SnapTS.desc)
+
+    val bronzeWarehouseSnapLatest = bronze_warehouse_snap.asDF
+      .withColumn("rnk", rank().over(lastWarehouseSnapW))
```
gueniai, line 1366: I see we do this in other modules as well, why do we use both rank as well as row_number instead of just row_number?

gueniai (changes requested):
Left you some minor comments

src/main/scala/com/databricks/labs/overwatch/pipeline/DbsqlTransforms.scala
```diff
@@ -0,0 +1,191 @@
...
+        PipelineFunctions.fillForward("spot_instance_policy",warehouse_name_gen_w),
+        PipelineFunctions.fillForward("enable_photon",warehouse_name_gen_w),
+        PipelineFunctions.fillForward("channel",warehouse_name_gen_w),
+        PipelineFunctions.fillForward("tags",warehouse_name_gen_w),
+        PipelineFunctions.fillForward("enable_serverless_compute",warehouse_name_gen_w),
+        PipelineFunctions.fillForward("warehouse_type",warehouse_name_gen_w)
+      )
+
+    warehouseRaw
+//    val warehouseWithStructs = warehouseRaw
```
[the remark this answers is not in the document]
aman-db: done

src/main/scala/com/databricks/labs/overwatch/pipeline/DbsqlTransforms.scala
```diff
@@ -0,0 +1,191 @@
...
+//      .withColumn("tags", SchemaTools.structFromJson(spark, warehouseRaw, "tags"))
+//      .scrubSchema
+//
+//    warehouseWithStructs
+//      .withColumn("tags", SchemaTools.structToMap(warehouseWithStructs, "tags"))
+  }
+
+  /**
+   * warehouseBaseFilled - if first run, baseline warehouse spec for existing warehouses that haven't been edited since
```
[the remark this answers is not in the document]
aman-db: done

src/main/scala/com/databricks/labs/overwatch/pipeline/Schema.scala
aman-db: done

src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
aman-db: as discussed we will leave this for now, until we get an answer/logic behind this ;)

src/main/scala/com/databricks/labs/overwatch/pipeline/Schema.scala
aman-db: Overwatch_RunID and PipelineSnapTS will be added after the execution of transformation function, it is being added to the tables using append function. Making it false will fail the module

gueniai (approved):
lgtm!

src/test/scala/com/databricks/labs/overwatch/utils/StringExtTest.scala
```diff
@@ -1,8 +1,10 @@
 package com.databricks.labs.overwatch.utils
 
 import com.databricks.labs.overwatch.utils.StringExt.StringHelpers
+import org.scalatest.Ignore
```
gueniai, line 4: Did you mean to make these changes?

gueniai: yes. There's the import and a commented out //@Ignore. Not sure if these are leftover from the development?

aman-db: @gueniai done

PR #675, opened 2023-01-03
0711 Release Staging

### LINKED ISSUE #115: Improve acquisition of Cloud Provider and OrgID
https://github.com/databrickslabs/overwatch/pull/114/files/9c064a62cc046c848baa580b09389b81bd5b7f43#diff-7c4309a6fba74925b551ba489878977bb3fa11aa15d9d6ee1186d90fab59f3f0

Likely can obtain orgid and cloud provider from spark tags, just needs testing. Likely better than using notebook context.

`spark.databricks.clusterUsageTags.clusterOwnerOrgId`
`spark.databricks.clusterUsageTags.cloudProvider`
`spark.databricks.clusterUsageTags.clusterAllTags`

May feed GCP Epic #61

### LINKED ISSUE #311: Details about Instance Profile
I am looking for details about  Instance Profile . In what table we can find them ?

### LINKED ISSUE #357: Overwatch On Photon - Broadcast exchange perf issue
Overwatch job may fail when being executed using the Photon engine. The spark_events_bronze module can fail due to a photonBroadcast exception. 

Proposed solution, add a hint to the df join between auditLogs and clusters_bronze to force a shuffle_hash join and avoid the timeout

`df1.join(df2.hint("SHUFFLE_HASH"), ...)`

### LINKED ISSUE #431: [FEAT] - Add bronze snapshots libraries / clusterPolicies
Create bronze snapshot tables for libraries and clusterPolicies

### LINKED ISSUE #433: [FEAT] -- Add additional snapshots to bronze

### LINKED ISSUE #567: [FEAT] Make the parallel API calls more generalize.

### LINKED ISSUE #646: column cluster_type is missing from the view jobruncostpotentialfact

### LINKED ISSUE #660: Serverless = HC Naming Convention
Databricks recently changed Serverless name to mean databricks owned compute whereas it used to mean High Concurrency clusters. High concurrency clusters should be referred to as HC or High Concurrency not serverless

### LINKED ISSUE #685: [FEAT] - Multi-Workspace CSV to Delta
Maintaining the multi-workspace config as a CSV is challenging and time-consuming and introduces opportunities for annoying bugs in formatting / etc when users edit and re-upload the CSV.

To avoid these challenges, the user should upload a CSV but instead of continuing to use this CSV the CSV should be converted to a delta dataset (as a path, not a table) and this is what should be referenced in the input parameter instead of the path to the CSV.

This allows users to simply pass update statements to the delta table to make config changes.

### LINKED ISSUE #686: [BUG] - SparkEventsBronze - Schema Exception
In certain circumstances where the sparkEventLogs have both executorAdded/Removed events and one of the following events with root level fields "Executor ID" and "executorId" respectively. This is not handled by the schemaScrubber and therefore needs to be manually handled.

Error is seen by the user in the pipReport as
`com.databricks.sql.transaction.tahoe.DeltaAnalysisException: Found duplicate column(s) in the data to save: executorid`

Additional Event types that can cause this issue
```
org.apache.spark.scheduler.SparkListenerExecutorBlacklistedForStage
org.apache.spark.scheduler.SparkListenerExecutorExcludedForStage
org.apache.spark.scheduler.SparkListenerExecutorUnblacklisted
org.apache.spark.scheduler.SparkListenerExecutorUnexcluded
```

### LINKED ISSUE #687: [FEAT] - APICallV2 - ReEnable support for .asStrings
.asStrings was possible in [ApiCall](https://github.com/databrickslabs/overwatch/blob/2b5e65334b802b960e69fe3e0bf16ad5d2a6d024/src/main/scala/com/databricks/labs/overwatch/ApiCall.scala#L111) but was removed in ApiCallV2. Please add support for this back as it has many valuable use cases.

Thanks.

### LINKED ISSUE #688: [BUG] -Overwatch MWS multithreaded run each thread should have its own SparkSession.
Now In overwatch multithreaded run all the threads are sharing a single spark session. So changing a config for a single table in sparkSession is impacting other pipeline tables. Hence each thread in Overwatch multi workspace deployment should have its own spark session.

### LINKED ISSUE #689: handling of trailing slashes in api_urls
**Overwatch Version**
0.7.1.0

**Describe the bug**
If you use the MultiWorkspaceRunner and the api_url contains trailing slash like `https://oregon.cloud.databricks.com/` instead of `https://oregon.cloud.databricks.com` then many targets will silently fail but the job will pass. 

It's easy to reproduce the error that causes the targets to fail in isolation in a notebook with

```
import com.databricks.labs.overwatch.ApiCallV2
import com.databricks.labs.overwatch.utils.ApiEnv

val jobsEndpoint = "jobs/list"

val query = Map(
  "limit" -> "25",
  "expand_tasks" -> "true",
  "offset" -> "0"
)


val apiEnv = ApiEnv(
  isLocal = false, 
  workspaceURL = "https://foo.cloud.databricks.com/",
  rawToken = "dapi_blah_blah",
  packageVersion = "0.7.1.0",
)

ApiCallV2(apiEnv, jobsEndpoint, query, 2.1)
  .execute()
  .asDF()
  .show()


```

Changing [this line](https://github.com/databrickslabs/overwatch/blob/2b5e65334b802b960e69fe3e0bf16ad5d2a6d024/src/main/scala/com/databricks/labs/overwatch/ApiMeta.scala#L103) to 

```
var request = Http(s"""${apiEnv.workspaceURL.stripSuffix("/")}/${apiV}/${apiName}""")
```

Would remedy it. Also would very much appreciate it if failures like this would fail the job, and not skip over the failures

### LINKED ISSUE #690: MultiWorkspace Concurrent Writes - Improve write lock logic
Delta does not support current writes when schema evolution takes place. 

Currently, OW defends against this with rudimentary sleeps and retries.

This should be improved by creating a global target table locking system only allowing targets to be written to when they are unlocked.

Suggest implementing a LinkedHasSet into global SparkSession object and if target fullTableName is present, the table should be considered locked. When the database.writeWithRetry is called this should be checked. If locked, recheck every 10s or something and the write should wait for this lock to be removed before continuing.

### LINKED ISSUE #693: [FEAT] - Par Workspace - Output Buffer - Output User Messages
Each module captures STDOUT for the user reference. This output is currently thrown all over the place since all workspaces and modules are executing simultaneously. I think a module needs an stdoutput buffer that only writes the output at the beginning and end of the module rather than at random times.

### LINKED ISSUE #694: Support cluster logging with 50+ mount points for MWS deployment
Issue: In multi workspace deployment if a remote workspace has more than 50 mount points then there is no guarantee that the cluster logs will be captured as part of Overwatch deployment.

Solution: Team will be creating a mount csv file which will be containing all the source of the mount point for the remote workspace .This path of the csv will be provided in the multiworkspace config file.

COMPLETE Docs for this process / feature

### LINKED ISSUE #700: [BUG] scope dbsql missing in Multiworkspace deployment
In Multiworkspace deployment scope dbsql is missing.

### LINKED ISSUE #701: [BUG] : MWS driver workspace performing dbfs/search-mounts api calls
Issue:
In multiworkspace deployment for getting the cluster logs we perform dbfs/search-mounts api calls to get the source informations of the mount points. This has to be done but not for all the workspaces. We should not perform dbfs/search-mounts api call for driver workspace because al the information regarding mount point for driver workspace is already available in driver workspace.

### LINKED ISSUE #704: [Bug] - Events are getting missed in Cluster_Spec_silver.
In some cases the events which are getting populated in audit_log_bronze getting missed in cluster_Spec_Silver
213189939-8b0869f9-09aa-434c-b408-15fa5ea14fb3

### LINKED ISSUE #714: Disable mount point check for AWS
Issue:
AWS don't have mount points to store cluster logs.But overwatch validation still checks the number of mount points for AWS.

Solution:
Please disable mount point check for AWS.

### LINKED ISSUE #715: [BUG] - Case condition in Excluded Scope config
Hi Team,

If any workspace doesn't have jobs then we need to exclude the scope in the multi workspace input config.
For example, I've mentioned "Jobs:pools" in the excluded scopes column and ran the pipeline.
But it ran for jobs modules and ended up with errors, because workspace don't have jobs.
Then changed it to lower case letters and it worked.

image

It should be like below

image

### LINKED ISSUE #727: JobRunsList - Enable StoreInTempLocation
The array of serialized results is getting too large when loading only from memory. `setStoreInTempLocation` should be enabled to enable batching to temp and not loading all results from memory.

Full Error Stack Trace:
```
Job aborted due to stage failure: Serialized task 136835:0 was 410720946 bytes, which exceeds max allowed: spark.rpc.message.maxSize (268435456 bytes). Consider increasing spark.rpc.message.maxSize or using broadcast variables for large values.
org.apache.spark.SparkException: Job aborted due to stage failure: Serialized task 136835:0 was 410720946 bytes, which exceeds max allowed: spark.rpc.message.maxSize (268435456 bytes). Consider increasing spark.rpc.message.maxSize or using broadcast variables for large values.
	at org.apache.spark.scheduler.DAGScheduler.failJobAndIndependentStages(DAGScheduler.scala:3312)
	at org.apache.spark.scheduler.DAGScheduler.$anonfun$abortStage$2(DAGScheduler.scala:3244)
	at org.apache.spark.scheduler.DAGScheduler.$anonfun$abortStage$2$adapted(DAGScheduler.scala:3235)
	at scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:62)
	at scala.collection.mutable.ResizableArray.foreach$(ResizableArray.scala:55)
	at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:49)
	at org.apache.spark.scheduler.DAGScheduler.abortStage(DAGScheduler.scala:3235)
	at org.apache.spark.scheduler.DAGScheduler.$anonfun$handleTaskSetFailed$1(DAGScheduler.scala:1424)
	at org.apache.spark.scheduler.DAGScheduler.$anonfun$handleTaskSetFailed$1$adapted(DAGScheduler.scala:1424)
	at scala.Option.foreach(Option.scala:407)
	at org.apache.spark.scheduler.DAGScheduler.handleTaskSetFailed(DAGScheduler.scala:1424)
	at org.apache.spark.scheduler.DAGSchedulerEventProcessLoop.doOnReceive(DAGScheduler.scala:3524)
	at org.apache.spark.scheduler.DAGSchedulerEventProcessLoop.onReceive(DAGScheduler.scala:3462)
	at org.apache.spark.scheduler.DAGSchedulerEventProcessLoop.onReceive(DAGScheduler.scala:3450)
	at org.apache.spark.util.EventLoop$$anon$1.run(EventLoop.scala:51)
	at org.apache.spark.scheduler.DAGScheduler.$anonfun$runJob$1(DAGScheduler.scala:1169)
	at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)
	at com.databricks.spark.util.FrameProfiler$.record(FrameProfiler.scala:80)
	at org.apache.spark.scheduler.DAGScheduler.runJob(DAGScheduler.scala:1157)
	at org.apache.spark.SparkContext.runJobInternal(SparkContext.scala:2727)
	at org.apache.spark.SparkContext.runJob(SparkContext.scala:2710)
	at org.apache.spark.SparkContext.runJob(SparkContext.scala:2822)
	at org.apache.spark.sql.catalyst.json.JsonInferSchema.infer(JsonInferSchema.scala:116)
	at org.apache.spark.sql.execution.datasources.json.TextInputJsonDataSource$.$anonfun$inferFromDataset$1(JsonDataSource.scala:121)
	at org.apache.spark.sql.execution.SQLExecution$.withSQLConfPropagated(SQLExecution.scala:392)
	at org.apache.spark.sql.execution.datasources.json.TextInputJsonDataSource$.inferFromDataset(JsonDataSource.scala:121)
	at org.apache.spark.sql.DataFrameReader.$anonfun$json$6(DataFrameReader.scala:562)
	at scala.Option.getOrElse(Option.scala:189)
	at org.apache.spark.sql.DataFrameReader.json(DataFrameReader.scala:562)
	at com.databricks.labs.overwatch.ApiCallV2.asDF(ApiCallV2.scala:489)
	at com.databricks.labs.overwatch.env.Workspace.getJobRunsDF(Workspace.scala:361)
	at com.databricks.labs.overwatch.pipeline.Bronze.appendJobRunsProcess$lzycompute(Bronze.scala:217)
	at com.databricks.labs.overwatch.pipeline.Bronze.appendJobRunsProcess(Bronze.scala:216)
	at com.databricks.labs.overwatch.pipeline.Bronze.$anonfun$executeModules$1(Bronze.scala:258)
	at scala.collection.immutable.List.foreach(List.scala:431)
	at com.databricks.labs.overwatch.pipeline.Bronze.executeModules(Bronze.scala:242)
	at com.databricks.labs.overwatch.pipeline.Bronze.run(Bronze.scala:280)
	at com.databricks.labs.overwatch.MultiWorkspaceDeployment.startBronzeDeployment(MultiWorkspaceDeployment.scala:181)
	at com.databricks.labs.overwatch.MultiWorkspaceDeployment.$anonfun$deploy$4(MultiWorkspaceDeployment.scala:390)
	at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)
	at scala.concurrent.Future$.$anonfun$apply$1(Future.scala:659)
	at scala.util.Success.$anonfun$map$1(Try.scala:255)
	at scala.util.Success.map(Try.scala:213)
	at scala.concurrent.Future.$anonfun$map$1(Future.scala:292)
	at scala.concurrent.impl.Promise.liftedTree1$1(Promise.scala:33)
	at scala.concurrent.impl.Promise.$anonfun$transform$1(Promise.scala:33)
	at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:64)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:750)
```

### LINKED ISSUE #729: [BUG] - Missing Cluster Events - AuditLog Delivery Delay results in missing some cluster events
Assume 2 runs, run1 and run2. 
t1, t2, tn represent linear time stamps
Assume te == time of event (i.e. timeevent)

Run1 goes from t1 until t2
Run2 goes from t2 until t3

HOWEVER 

Create event occurred at te but was not delivered via audit log pipeline until after t2; thus during Run1 clusterEvents did not look for events on that cluster as Overwatch didn't know this cluster was present (i.e. Overwatch did not have this ID yet). Run2 DOES see the clusterId and tries to grab the events for the cluster; however, the events associated with this cluster are ACTUALLY between t1 and t2 but the API Query is for events between t2 and t3; thus missing all events for the cluster between t1 and t2.

IMPACT:
Missing clusterEvents in bronze is very bad as it results in inaccurate costs and utilization downstream. Furthermore, since this is bronze and the clusterEvents api history has limited retention, there's no way to recover these after some time.

NOTE:
This only happens for new clusters, typically clusters that were created near the end of the pipeline run time (i.e. near, but before t2). As a result, high-frequency jobs clusters are the most likely to experience this issue.

PROPOSED SOLUTION:
Override the fromTime of the clusterEvents API query from t1 to t1 - 1 day (audit log delivery SLA is 24h). This will result in duplicate events; as such clusterEvents Bronze must be switched from append mode to merge mode. More specifically, merge whenNotMatched Insert (no update whenMatched). This will ensure that only new events get landed but also that no events will be missed.

### LINKED ISSUE #731: [BUG] Jobs_Snap_Bronze - new_cluster not cleansed
the new_cluster field in jobs_snapshot_bronze is not cleansed and has structs in it where it should have maps.

Error will present itself in Silver JobStatusSilver as the following. This is because on first run a union is done between jobs_snapshot_bronze and the derived job_status_silver to initialize all jobs. During the union, the `new_cluster` field cannot be merged since jobs_snapshot_bronze wasn't cleansed.

This ONLY occurs on first run and only when customer has legacy single-task jobs created with jobs api 2.0

```
FAILED --> 
ERROR:
Silver_JobsStatus FAILED -->

Failed to merge fields 'new_cluster' and 'new_cluster'. Failed to merge fields 'aws_attributes' and 'aws_attributes'. Failed to merge incompatible data types MapType(StringType,StringType,true) and StructType(StructField(availability,StringType,true),StructField(first_on_demand,LongType,true),StructField(zone_id,StringType,true))
org.apache.spark.sql.AnalysisException: Failed to merge fields 'new_cluster' and 'new_cluster'. Failed to merge fields 'aws_attributes' and 'aws_attributes'. Failed to merge incompatible data types MapType(StringType,StringType,true) and StructType(StructField(availability,StringType,true),StructField(first_on_demand,LongType,true),StructField(zone_id,StringType,true))

```

### LINKED ISSUE #737: [BUG] - AWS - Missing Some Day Audit Events
Dates glob was designed to be fromDate (inclusive) untilDate (exclusive) but this will drop all current day events from audit logs in AWS when only running once per day.

Originally designed this way due to compatibility with audit logs v1 but audit logs v2.2 is out now and an updated method is required.

Switching to the following
fromDate (inclusive) untilDate (INCLUSIVE)

To ensure no duplicate audit logs and audit log hash will be generated for each record and a whenNotMatched then Insert will be used to ensure audit logs don't get duplicated on ingest.

### LINKED ISSUE #748: [BUG] - MultiWorkspace - Silver / Gold Not starting after bronze per workspace
When Bronze is finished for workspace 1, it should continue on to silver even if workspaces 2-n are still processing Bronze. Currently, Silver waits for Bronze to complete all workspaces before continuing.

Same is true for gold waiting on silver.

### LINKED ISSUE #749: [PERF] - CLSF - FillMeta Causing Skew
CLSF needs to lookup metadata from cluster. When doing this it can cause significant skew since some clusters have many many more states than others.

### LINKED ISSUE #750: [FEAT] - getWorkspaceByDatabase - Enable functionality for remote workspaces
`getWorkspaceByDatabase` is often used to get a workspace object but after the deployment of multi-workspace deployments we forgot to enable this utility to work for instantiating remote workspaces.

### LINKED ISSUE #757: [BUG] - DBU & Cost Calculation Incorrect (Primarily Photon Clusters)
The calculations for DBUs and DBU Cost are not correct.

### LINKED ISSUE #758: [BUG] - First Run - ClusterSpec - Missing Some Record Initialization
On first run, some of the missing audit log cluster ids may not be imputed when first run does not run through current timestamp due to snapshot filter.

### LINKED ISSUE #769: Mutli-Workspace Config Null handler
Often times, customers wind up passing "null" or "Null" into fields that should be "" empty string. I think we should auto-convert case-insenstitive "null" from the CSV to be an actual null so as to mitigate config issues.

The "Null" looks the same as a true null in the spark.read...config.csv thus making it more challenging to diagnose without downloading and reviewing the actual text of the CSV.

### LINKED ISSUE #771: Par Spark Conf - Replicate global configs to thread-specific new sessions
when thread-specific spark sessions are created the global conf should be mirrored into the new sparkSession and all global overrides should be combined into a single override location.

### LINKED ISSUE #772: MultiWorkspace Deployments - Pull workspace Instantiation outside of zone specific deployment
To ensure spark configs are passed properly and to optimize workloads workspace instantiation should only be done once per full pipeline set 

Follow BatchRunner for example -- workspace is instantiated once and then passed into each of the pipelines not once per pipeline. This is important to ensure the initialized configs (including spark configs) are consistent throughout.

The first step inside the future should be to instantiate the workspace from the validated params. This should be moved up outsize of `startBronzeDeployment`, `startSilverDeployment`, `startGoldDeployment`

Example:
```scala
implicit val ec: ExecutionContextExecutor = ExecutionContext.fromExecutor(Executors.newFixedThreadPool(parallelism))
  params.foreach(deploymentParams => {
    val future = Future {
    val workspace = Initializer(multiWorkspaceParams.args, debugFlag = false,
      apiURL = Some(multiWorkspaceParams.apiUrl),
      organizationID = Some(multiWorkspaceParams.workspaceId))

      if (zones.toLowerCase().contains("bronze")) {
        startBronzeDeployment(workspace)
      }
      if (zones.toLowerCase().contains("silver")) {
        startSilverDeployment(workspace)
      }
      if (zones.toLowerCase().contains("gold")) {
        startGoldDeployment(workspace)
      }
    }
```

Then inside of each of the starts, use `workspace.getConfig` to access configs from the workspace object.

Required for #771 to function properly.

### LINKED ISSUE #775: mount_mapping_path validation
Need to have validation for mount_mapping_path. Validations can be as below
1)File should exist in the provided path.
2)Expected column should exist on provided csv.
3)If mount_mapping_path is provided then we should not perform api call for mount-point mapping.

GeekSheikh:
closes #115 
closes #357 
closes #431 
closes #433 
closes #567 
closes #646 
closes #660 
closes #688 
closes #685 
closes #686 
closes #687 
closes #689 
closes #690 
closes #693 
closes #694 
closes #700 
closes #701 
closes #704 
closes #714 
closes #715 
closes #727 
closes #729 
closes #731 
closes #737 
closes #748 
closes #749 
closes #750 
closes #757 
closes #758 
closes #769 
closes #771 
closes #772 
closes #775 

partially resolves #311


alexott (approved):

brij-raghuwanshi-db (commented):
Looks Good.

PR #678, opened 2023-01-04
Spark events bronze par session

### LINKED ISSUE #686: [BUG] - SparkEventsBronze - Schema Exception

### LINKED ISSUE #688: [BUG] -Overwatch MWS multithreaded run each thread should have its own SparkSession.

sriram251-code:
closes #686 
closes #688


src/main/scala/com/databricks/labs/overwatch/ApiCallV2.scala
```diff
@@ -607,7 +600,6 @@ class ApiCallV2(apiEnv: ApiEnv) extends SparkSessionWrapper {
       responseCodeHandler(response)
       _apiResponseArray.add(response.body)
       if (apiMeta.storeInTempLocation && successTempPath.nonEmpty) {
-        accumulator.add(1)
```
GeekSheikh, line 610: Why are you removing this? Don't we still need to increment the accumulator?

src/main/scala/com/databricks/labs/overwatch/MultiWorkspaceDeployment.scala
```diff
@@ -239,6 +246,9 @@ class MultiWorkspaceDeployment extends SparkSessionWrapper {
           fullMsg,
           Some(multiWorkspaceParams.deploymentId)
         ))
+    }finally {
+        SparkSessionWrapper.sessionsMap.remove(Thread.currentThread().getId)
+        logger.log(Level.INFO, s"""Removed ${Thread.currentThread().getId} from sessionMap""")
```
GeekSheikh, line 251: Instead of triplicating this logic, please create a function that takes in the threadID and completes this logic. This func should probably reside inside of SparkSessionWrapper.

src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala
```diff
@@ -741,7 +741,7 @@ trait BronzeTransforms extends SparkSessionWrapper {
       val pathsGlob = validNewFilesWMetaDF
         .filter(!'failed && 'withinSpecifiedTimeRange)
         .orderBy('fileSize.desc)
-        .select('fileName)
+        .select('filename)
```
GeekSheikh, line 744: I'm not sure how you got this change. It's a good change and one I was planning to implement but curious how you got it. Please leave it for case sensitivity reasons.

src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala
```diff
@@ -798,6 +798,7 @@ trait BronzeTransforms extends SparkSessionWrapper {
             TransformFunctions.stringTsToUnixMillis('timestamp)
           } else col("Timestamp")
 
+          logger.log(Level.INFO, s"DEBUG: SparkEventsBronze Case Sensitivity set to: ${spark.conf.get("spark.sql.caseSensitive")}")
```
GeekSheikh, line 801: Do we want to leave this log? Ok if so, just wanted to make sure you did it on purpose. If you're going to leave it, we probably need to add the orgID in context to the message so we know for which orgID the conf is being set.

src/main/scala/com/databricks/labs/overwatch/pipeline/Pipeline.scala
```diff
@@ -384,8 +384,9 @@ class Pipeline(
     dbutils.fs.rm(config.tempWorkingDir)
 
     postProcessor.refreshPipReportView(pipelineStateViewTarget)
-
-    spark.catalog.clearCache()
+    //TODO clearcache will clear global cache multithread performance issue
+   // spark.catalog.clearCache()
```
GeekSheikh, line 388: REMINDER - We need to address this before merging.

src/main/scala/com/databricks/labs/overwatch/utils/SparkSessionWrapper.scala
```diff
@@ -1,9 +1,21 @@
...
 import org.apache.spark.SparkContext
 import org.apache.spark.sql.SparkSession
 
+import java.util.concurrent.ConcurrentHashMap
+import scala.collection.JavaConverters._
+
+
+object SparkSessionWrapper {
+
+   var parSessionsOn = false
```
GeekSheikh, line 14: please convert this var to private and create getter/setters.

Why did we wind up using this instead of the spark.conf("overwatch.parSessions")? I'm wondering if maintaining the global state of yet another var may cause other issues. Really no diff than the global sparkconf but just curious why we switched the approach.

src/main/scala/com/databricks/labs/overwatch/utils/SparkSessionWrapper.scala
```diff
@@ -20,30 +33,38 @@ trait SparkSessionWrapper extends Serializable {
   @transient
```
GeekSheikh, line 33: this @transient should be removed. I know you didn't add it but not sure how it got here. Probably didn't get moved with the spark val like it should have in some earlier PR

src/main/scala/com/databricks/labs/overwatch/utils/SparkSessionWrapper.scala
```diff
@@ -20,30 +33,38 @@ trait SparkSessionWrapper extends Serializable {
...
   lazy protected val _envInit: Boolean = envInit()
 
+
+  protected def buildSpark(): SparkSession = {
```
GeekSheikh, line 37: I think this should be private, right?

src/main/scala/com/databricks/labs/overwatch/utils/SparkSessionWrapper.scala
```diff
@@ -20,30 +33,38 @@ trait SparkSessionWrapper extends Serializable {
...
-      .master("local[*]")
-      .config("spark.driver.maxResultSize", "8g")
-      .appName("OverwatchBatch")
-//    Useful configs for local spark configs and/or using labs/spark-local-execution
-//    https://github.com/databricks-academy/spark-local-execution
-//      .config("spark.driver.bindAddress", "0.0.0.0")
-//      .enableHiveSupport()
-//      .config("spark.warehouse.dir", "metastore")
-      .getOrCreate()
+  def spark(globalSession : Boolean = false): SparkSession = {
```
GeekSheikh, line 50: This should be private[overwatch] -- I don't want people using this outside of the package, I want to ensure they use val spark

src/main/scala/com/databricks/labs/overwatch/utils/SparkSessionWrapper.scala
sriram251-code: done. also noticed that _envInit is never being used.

src/main/scala/com/databricks/labs/overwatch/utils/SparkSessionWrapper.scala
sriram251-code: yes done.

src/main/scala/com/databricks/labs/overwatch/utils/SparkSessionWrapper.scala
sriram251-code: done

src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala
GeekSheikh: this needs to be filename all lower case, switched it back to all lower case.

src/main/scala/com/databricks/labs/overwatch/pipeline/Pipeline.scala
GeekSheikh: created #722

GeekSheikh (approved):
Looks like this will work, will do complete testing on the release branch.

PR #679, opened 2023-01-04
SBT update

carwilki:
PR Updates sbt to 1.8.1. Removes the courser plugin as it is now bundled with sbt and is the default dependency resolver. Also with the changes the project is now bloop compatible.


GeekSheikh:
Thank you for your PR @carwilki -- we plan to integrate this into version 0.7.1.1 release. Thanks @alexott for your review.

Issue #404, opened 2022-05-20
Upgrade - Simplify Process and Code


GeekSheikh:
When schema changes occur an upgrade to the schema is required.

It's currently challenging to upgrade across version
There's a lot of redundant code in the Upgrade package
- refactor the error report to make it easier to use
- refactor startFrom step logic as necessary to allow upgrade resume / recovery
- create upgrade chains that allow for upgrades across versions (i.e. from 0506 to 062 for example)

GeekSheikh:
@sudharshanraja-db -- please provide a status update. Will this be complete in January?

sudharshanraja-db:
HI @GeekSheikh  Yes this weekend i spent some time understanding the code flow ,
Next week i  have   bandwidth  to work on this.

gueniai:
This issue will be a P1 to implement in 0.7.2.1, it will be a P0 if it slips to 0.7.3.0

PR #681, opened 2023-01-07
Ov 404 simplify upgrade process v1

sudharshanraja-db:


src/main/scala/com/databricks/labs/overwatch/utils/UpgradeV2.scala
```diff
@@ -0,0 +1,405 @@
...
+class UpgradeTo0610(prodWorkspace: Workspace, startStep: Int = 1, endStep: Int = 3, enableUpgradeBelowDBR104: Boolean = false) extends UpgradeHandler {
+
+
+  def this(workspace: Workspace) {
+    this(workspace, 1, 3,false)
+  }
+
+
+  override def upgrade(): DataFrame = {
+    val tempDir = s"/tmp/overwatch/upgrade0610_status__ctrl_0x111/${System.currentTimeMillis()}"
```
GeekSheikh, line 28: this prefix should taken from `prodWorkspace.getConfig.tempWorkingDir` by default but be able to be optionally overridden in the constructor. Please default it to null string so user doesn't have to pass it in.

src/main/scala/com/databricks/labs/overwatch/utils/UpgradeV2.scala
```diff
@@ -0,0 +1,405 @@
...
+    validateNumericalSchemaVersion(numericalSchemaVersion, 600, 610)
+    val upgradeStatus: ArrayBuffer[UpgradeReport] = ArrayBuffer()
+    val dbrVersion = spark.conf.get("spark.databricks.clusterUsageTags.effectiveSparkVersion")
+    val dbrMajorV = dbrVersion.split("\\.").head
+    val dbrMinorV = dbrVersion.split("\\.")(1)
+    val dbrVersionNumerical = s"$dbrMajorV.$dbrMinorV".toDouble
+    val initialSourceVersions: concurrent.Map[String, Long] = new ConcurrentHashMap[String, Long]().asScala
+    val packageVersion = blankConfig.getClass.getPackage.getImplementationVersion.replaceAll("\\.", "").tail.toInt
+    val startingSchemaVersion = SchemaTools.getSchemaVersion(etlDatabaseName).split("\\.").takeRight(1).head.toInt
+    validateSchemaAndPackageVersion(startingSchemaVersion, packageVersion, 600, 610)
```
GeekSheikh, line 45: Most of these items are common across all upgrades such as source and target version etc. The only exception is the dbr version stuff for the 0610 upgrade. All of the vals that are derived, can they be moved to the abstract class so they don't have to be defined every time?

src/main/scala/com/databricks/labs/overwatch/utils/UpgradeV2.scala
```diff
@@ -0,0 +1,405 @@
...
+
+    (startStep to endStep).foreach {
+      case 1 =>
+        val stepMsg = Some("Step 1: Upgrade Schema - Job Status Silver")
```
GeekSheikh, line 49: I want to try and clean this up so that all upgrades ultimately submit a map of step number and upgradeStepFunction to the UpgradeHandler. Hopefully, it can look something like

```scala
private def 0610Step1(...) = ???
private def 0610Step2(...) = ???

val upgradeMap(stepNumber -> upgradeFunc, ...)
upgrade(upgradeMap)
```
something like this. The above is just pseudo code obviously but hopefully that is clear enough. The idea being that the dev creates n functions and then passes those to the upgrade -- I think this will make the code much cleaner and easier to navigate, thoughts?

GeekSheikh: Inside the upgrade handler all the exception handlers should be obfuscated there meaning all the try / catch / finally stuff should be in the abstract class or executor not in the actual upgrade script.

src/main/scala/com/databricks/labs/overwatch/utils/UpgradeV2.scala
```diff
@@ -0,0 +1,405 @@
...
+        println(stepMsg.get)
+        logger.log(Level.INFO, stepMsg.get)
+        val targetName = "job_status_silver"
+        try {
+          checkIfTargetExists(etlDatabaseName, targetName)
+          initialSourceVersions.put(targetName, Helpers.getLatestTableVersionByName(s"${etlDatabaseName}.${targetName}"))
+          val jobSilverDF = spark.table(s"${etlDatabaseName}.${targetName}")
+          removeNestedColumnsAndSaveAsTable(jobSilverDF,"new_settings", Array("tasks", "job_clusters"),etlDatabaseName,targetName)
```
GeekSheikh, line 57: This should be the entirety of the step 1 upgrade func for 0610 upgrade. This hopefully will be all the code necessary to complete the 0610 step 1 upgrade as a dev building the upgrade.

src/main/scala/com/databricks/labs/overwatch/utils/UpgradeExecutor.scala
```diff
@@ -0,0 +1,72 @@
+package com.databricks.labs.overwatch.utils
+
+import com.databricks.labs.overwatch.env.Workspace
+import org.apache.spark.sql.DataFrame
+
+import scala.collection.mutable
+
+class UpgradeExecutor(val etldbname: String, val fromVersion: Int, val toVersion: Int, val executeSingleStep: Int = -1) {
+
+  val versionMap = Map("600-609" -> classOf[UpgradeTo0610], "610-699" -> classOf[UpgradeTo0700])
```
GeekSheikh, line 10: Would you have a versionMap for each upgrade path? Is this just an example for 1 of them?

GeekSheikh (changes requested):

sudharshanraja-db:
closes #404

PR #684, opened 2023-01-09
adding new snapshots to bronze layer

aman-db:
* adding new snapshots to bronze layer

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/pipeline/Bronze.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/pipeline/Bronze.scala
@@ -185,12 +185,14 @@ class Bronze(_workspace: Workspace, _database: Database, _config: Config)
   lazy private[overwatch] val libsSnapshotModule = Module(1007, "Bronze_Libraries_Snapshot", this)
   lazy private val appendLibsProcess = ETLDefinition(
     workspace.getClusterLibraries,
+    Seq(cleanseRawLibsSnapDF),
     append(BronzeTargets.libsSnapshotTarget)
   )
 
   lazy private[overwatch] val policiesSnapshotModule = Module(1008, "Bronze_Policies_Snapshot", this)
   lazy private val appendPoliciesProcess = ETLDefinition(
     workspace.getClusterPolicies,
+    Seq(cleanseRawPoliciesSnapDF),
     append(BronzeTargets.policiesSnapshotTarget)
   )
 
@@ -215,6 +217,7 @@ class Bronze(_workspace: Workspace, _database: Database, _config: Config)
   lazy private[overwatch] val jobRunsSnapshotModule = Module(1012, "Bronze_Job_Runs_Snapshot", this) // check module number
   lazy private val appendJobRunsProcess = ETLDefinition(
     workspace.getJobRunsDF,
+    Seq(cleanseRawJobRunsSnapDF),
     append(BronzeTargets.jobRunsSnapshotTarget)
   )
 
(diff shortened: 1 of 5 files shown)


aman-db:
closes https://github.com/databrickslabs/overwatch/issues/431
closes https://github.com/databrickslabs/overwatch/issues/433

partially resolves https://github.com/databrickslabs/overwatch/issues/311

src/main/scala/com/databricks/labs/overwatch/pipeline/Bronze.scala
```diff
@@ -182,6 +182,42 @@ class Bronze(_workspace: Workspace, _database: Database, _config: Config)
...
+  )
+
+  lazy private[overwatch] val globalInitScSnapshotModule = Module(1011, "Bronze_Global_Init_Scripts_Snapshot", this)
+  lazy private val appendGlobalInitScProcess = ETLDefinition(
+    workspace.getGlobalInitScripts,
+    append(BronzeTargets.globalInitScSnapshotTarget)
+  )
+
+  lazy private[overwatch] val jobRunsSnapshotModule = Module(1012, "Bronze_Job_Runs_Snapshot", this) // check module number
+  lazy private val appendJobRunsProcess = ETLDefinition(
```
GeekSheikh, line 216: I noticed that none of these bronze lands have any transforms which may be fine for most (maybe all of them) but I want to make sure a schema analysis has been completed to ensure there aren't any user defined key/value pairs being landed as structs. If so, we need a bronzeTransform to cleanse. see `cleanseRawClusterSnapDF` func for example on cleansing bronze lands. There also may need to be a schema scrub if there are any spaces or special characters in any of the return schemas.

src/main/scala/com/databricks/labs/overwatch/pipeline/PipelineTargets.scala
```diff
@@ -154,6 +154,53 @@ abstract class PipelineTargets(config: Config) {
...
+    lazy private[overwatch] val instanceProfileSnapshotTarget: PipelineTable = PipelineTable(
+      name = "instance_profile_snapshot_bronze",
+      _keys = Array("cluster_id", "Overwatch_RunID"),
+      config,
+      incrementalColumns = Array("Pipeline_SnapTS"),
+      partitionBy = Seq("organization_id")
+    )
+
+    lazy private[overwatch] val tokenSnapshotTarget: PipelineTable = PipelineTable(
+      name = "token_snapshot_bronze",
```
GeekSheikh, line 182: please convert all single asset names to plural to maintain uniformity throughout bronze. This name should be tokens_snapshot_bronze instead of token.

instance_profile should be instance_profiles... etc. Notice how even in this change you have some plural (i.e. libs) and some singular. Make uniform to plural

GeekSheikh (changes requested):
Minor changes requested, please also complete pr proof notebook.

src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala
```diff
@@ -1022,4 +1022,29 @@ trait BronzeTransforms extends SparkSessionWrapper {
...
+    outputDF
+      .withColumn("library_statuses", explode(col("library_statuses")))
+  }
+
+  protected def cleanseRawPoliciesSnapDF(df: DataFrame): DataFrame = {
+    val outputDF = SchemaScrubber.scrubSchema(df)
+
+    outputDF
+      .withColumn("definition", SchemaTools.structFromJson(spark,outputDF, "definition"))
+  }
```
GeekSheikh, line 1037: Not needed

GeekSheikh (approved):
Requires changes but doing them myself due to time crunch. Please see comments for more details and look at changes on the next commit. cc @aman-db

GeekSheikh:
The `tasks` field of jobRuns is still not expanding as expected. We need to keep an eye on that and ensure that if cluster details start to appear in that API that we add more cleansing work to to the jobRuns snap df.

PR #705, opened 2023-01-18
Overwatch on photon broadcast exchange perf issue

### LINKED ISSUE #357: Overwatch On Photon - Broadcast exchange perf issue

souravbaner-da:
Closes #357


GeekSheikh (approved):

brij-raghuwanshi-db (commented):
@souravbaner-da :
Looks good.
It would be great if we get the scenario of what was not working fine before making this change.
Also, what is the impact of this change?
If we are able to document it, that will be great.

PR #713, opened 2023-01-27
Reading config from delta implemented.

### LINKED ISSUE #685: [FEAT] - Multi-Workspace CSV to Delta

### LINKED ISSUE #714: Disable mount point check for AWS

sriram251-code:
closes #685 
closes #714 
Reading config from delta implemented.
Skip mount point check for AWS added.

PR proof as below

https://adb-2222170229861029.9.azuredatabricks.net/?o=2222170229861029#notebook/3480796113320982/command/3480796113320991

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/MultiWorkspaceDeployment.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/MultiWorkspaceDeployment.scala
@@ -19,20 +19,20 @@ import scala.concurrent.{ExecutionContext, ExecutionContextExecutor, Future}
 /** *
  * MultiWorkspaceDeployment class is the main class which runs the deployment for multiple workspaces.
  *
- * @params configCsvPath: path of the csv file which will contain the different configs for the workspaces.
+ * @params configCsv: path of the csv file/Delta table which will contain the different configs for the workspaces.
  * @params tempOutputPath: location which will be used as a temp storage.It will be automatically cleaned after each run.
  * @params apiEnvConfig: configs related to api call.
  */
 object MultiWorkspaceDeployment extends SparkSessionWrapper {
 
 
-  def apply(configCsvPath: String): MultiWorkspaceDeployment = {
-    apply(configCsvPath, "/mnt/tmp/overwatch")
+  def apply(configCsv: String): MultiWorkspaceDeployment = {
+    apply(configCsv, "/mnt/tmp/overwatch")
   }
 
-  def apply(configCsvPath: String, tempOutputPath: String) = {
+  def apply(configCsv: String, tempOutputPath: String) = {
     new MultiWorkspaceDeployment()
-      .setConfigCsvPath(configCsvPath)
+      .setConfigCsv(configCsv)
       .setOutputPath(tempOutputPath)
       .setPipelineSnapTime()
   }
@@ -45,7 +45,7 @@ class MultiWorkspaceDeployment extends SparkSessionWrapper {
 
   private val logger: Logger = Logger.getLogger(this.getClass)
 
-  private var _configCsvPath: String = _
+  private var _configCsv: String = _
 
   private var _outputPath: String = _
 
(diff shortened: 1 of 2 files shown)


src/main/scala/com/databricks/labs/overwatch/MultiWorkspaceDeployment.scala
```diff
@@ -313,17 +314,43 @@ class MultiWorkspaceDeployment extends SparkSessionWrapper {
     true
   }
 
+
+  private def generateBaseConfig(configCsvPath: String): DataFrame = {
```
GeekSheikh, line 318: should we change param name from configCSVPath to configPath or something since it's no longer required to be a csv? Wondering if we should just go ahead and make the switch in the deployment runner now too -- easier to change now than later.

src/main/scala/com/databricks/labs/overwatch/validation/DeploymentValidation.scala
```diff
@@ -100,52 +100,61 @@ object DeploymentValidation extends SparkSessionWrapper {
...
-         |DBPATWorkspaceScope:${conf.secret_scope}
-         |SecretKey_DBPAT:${conf.secret_key_dbpat}""".stripMargin
-    try {
-      val patToken = dbutils.secrets.get(scope = conf.secret_scope, key = conf.secret_key_dbpat)
-      val apiEnv = ApiEnv(false, conf.api_url, patToken, getClass.getPackage.getImplementationVersion)
-      val endPoint = "dbfs/search-mounts"
-      val mountCount = ApiCallV2(apiEnv, endPoint).execute().asDF().count()
-      if(mountCount<50)
-        {
+    if (conf.cloud.toLowerCase == "azure" &&  conf.workspace_id.trim != spark.conf.get("spark.databricks.clusterUsageTags.clusterOwnerOrgId")) {
```
GeekSheikh, line 103: can we use the getOrgID func instead of duplicating this critical logic?

GeekSheikh, line 103: Please add some docs around this if statement and even the func would be helpful.

I think it would also be helpful to convert the section below into a variable that is more english legible such as 
```scala 
val isAzure = conf.cloud.toLowerCase == "azure"
val isRemoteWorkspace = conf.workspace_id.trim != Initializer.getOrgId
if (isAzure && isRemoteWorkspace) ...
```
I think it makes the if much more legible.
```scala
conf.workspace_id.trim != spark.conf.get("spark.databricks.clusterUsageTags.clusterOwnerOrgId"
```

GeekSheikh: shouldn't there be one more case added to the if statement as well? and mount_mapping_path is null or something like that to disable this validation if a mount_mapping_path was added?

Note: This should only be disabled if the mount_mapping_path is provided AND has org_id for the conf.workspace_id

GeekSheikh (changes requested):

src/main/scala/com/databricks/labs/overwatch/validation/DeploymentValidation.scala
sriram251-code: implemented for mount_mapping_path validation will implement in a different branch once this branch is merged

src/main/scala/com/databricks/labs/overwatch/MultiWorkspaceDeployment.scala
sriram251-code: done

src/main/scala/com/databricks/labs/overwatch/validation/DeploymentValidation.scala
sriram251-code: done

sriram251-code: for mount_mapping_path validation i will have it in a different branch once this PR is merged

GeekSheikh (approved):
outstanding comments to be handled in subsequent PR including tests and code comments.

PR #717, opened 2023-01-31
shuffle partition changed to String

### LINKED ISSUE #711: Runner can not handle 'auto' partition setting
Overwatch currently assumes that `spark.sql.shuffle.partitions` is set to an `Int` and will fail to run otherwise.

https://github.com/databrickslabs/overwatch/blob/2b5e65334b802b960e69fe3e0bf16ad5d2a6d024/src/main/scala/com/databricks/labs/overwatch/pipeline/Initializer.scala#L581

Databricks runtime currently supports setting partitions to `auto` which then fails overwatch jobs.

Code should likely either be updated to handle `auto` or documentation updated to explicitly call out this issue and recommend against setting `auto` partitions.

sriram251-code:
closes #711

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/pipeline/Initializer.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/pipeline/Initializer.scala
@@ -578,7 +578,7 @@ object Initializer extends SparkSessionWrapper {
     }
     config.registerInitialSparkConf(spark.conf.getAll)
     config.setInitialWorkerCount(getNumberOfWorkerNodes)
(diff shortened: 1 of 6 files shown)


GeekSheikh (changes requested):
MISSED this one -- needs special case if it's set to auto since it cannot be multiplied by shuffleFactor
https://github.com/databrickslabs/overwatch/blob/2b5e65334b802b960e69fe3e0bf16ad5d2a6d024/src/main/scala/com/databricks/labs/overwatch/pipeline/Module.scala#L60

sriram251-code:
done

PR #738, opened 2023-02-14
auditRaw - mergeInsertOnly

### LINKED ISSUE #737: [BUG] - AWS - Missing Some Day Audit Events

GeekSheikh:
closes #737


GeekSheikh:
branched from `729_missing_cluster_events` which is to be merged in #730 so better to review this after #730 is merged to minimize changes to review.

sriram251-code (approved):

PR #740, opened 2023-02-15
excluded scope enhanced

sriram251-code:


src/main/scala/com/databricks/labs/overwatch/MultiWorkspaceDeployment.scala
```diff
@@ -118,7 +118,7 @@ class MultiWorkspaceDeployment extends SparkSessionWrapper {
       val customWorkspaceName: String = config.workspace_name
       val standardScopes = "audit,sparkEvents,jobs,clusters,clusterEvents,notebooks,pools,accounts,dbsql".split(",").toBuffer
       if (config.excluded_scopes != null) {
-        config.excluded_scopes.split(":").foreach(scope => standardScopes -= scope)
+        config.excluded_scopes.toLowerCase().split(":").foreach(scope => standardScopes -= scope)
```
GeekSheikh, line 121: so I notice that you're setting excluded_scopes to lower case but the `standardScopes` are not all lower case so do you have any matching anywhere or can we confirm that if you change excluded scopes to lower case that it will exclude non-lower-case scopes?

EX: Is this a problem?
sparkEvents != sparkevents

GeekSheikh (changes requested):

PR #747, opened 2023-02-15
567 parallel api calls

aman-db:
Generalised parallel api calls for - 
* sqlQueryHistory
* landClusterEvents

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/ApiCallV2.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/ApiCallV2.scala
@@ -633,22 +633,27 @@ class ApiCallV2(apiEnv: ApiEnv) extends SparkSessionWrapper {
    * @return
    */
   def makeParallelApiCalls(endpoint: String, jsonInput: Map[String, String], config: Config): String = {
-    var startValue = Integer.parseInt(jsonInput.get("start_value").get)
-    val endValue = Integer.parseInt(jsonInput.get("end_value").get)
-    val incrementCounter = Integer.parseInt(jsonInput.get("increment_counter").get)
-    val finalResponseCount = Integer.parseInt(jsonInput.get("final_response_count").get)
+    var startValue = jsonInput.get("start_value").get.toLong
+    val endValue = jsonInput.get("end_value").get.toLong
+    val incrementCounter = jsonInput.get("increment_counter").get.toLong
+    val finalResponseCount = jsonInput.get("final_response_count").get.toLong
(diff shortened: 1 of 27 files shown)


src/test/scala/com/databricks/labs/overwatch/utils/SchemaToolsTest.scala
```diff
@@ -91,6 +91,111 @@ class SchemaToolsTest extends AnyFunSpec with SparkSessionTestWrapper with Given
     }
   }
 
+  describe("nestedColExists") {
```
GeekSheikh, line 94: @aman-db -- is this supposed to be part of this PR?

src/test/scala/com/databricks/labs/overwatch/utils/StringExtTest.scala
```diff
@@ -1,10 +1,9 @@
 package com.databricks.labs.overwatch.utils
 
-import com.databricks.labs.overwatch.SparkSessionTestWrapper
 import com.databricks.labs.overwatch.utils.StringExt.StringHelpers
 import org.scalatest.funspec.AnyFunSpec
 
-class StringExtTest extends AnyFunSpec with SparkSessionTestWrapper {
+class StringExtTest extends AnyFunSpec {
```
GeekSheikh, line 6: @aman-db -- is this supposed to be part of this PR?

src/main/scala/com/databricks/labs/overwatch/ApiCallV2.scala
```diff
@@ -619,6 +625,124 @@ class ApiCallV2(apiEnv: ApiEnv) extends SparkSessionWrapper {
...
   }
 
+  /**
+   * Function to make parallel API calls. Currently this functions supports only SqlQueryHistory and ClusterEvents
+   * @param endpoint
+   * @param jsonInput
+   * @param config
+   * @return
+   */
+  def makeParallelApiCalls(endpoint: String, jsonInput: Map[String, String], config: Config): String = {
```
GeekSheikh, line 635: I really think this belongs in a class and that we should refactor APICall, APICallV2, and APIMeta and move them into a new namespace (folder) like API. Inside of API perhaps we create something like `APIParHandler`. There are so many variables at the top which look like class vars and then we could break up the future section into several smaller functions to simplify support for this. Thoughts? This really isn't me telling you to do this, looking for discussion / thoughts / feedback on this idea. @sriram251-code, I would like your feedback too.

GeekSheikh (changes requested):

PR #756, opened 2023-02-17
adding generic api calls function

### LINKED ISSUE #567: [FEAT] Make the parallel API calls more generalize.

aman-db:
closes #567

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/ApiCallV2.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/ApiCallV2.scala
@@ -1,5 +1,6 @@
 package com.databricks.labs.overwatch
 
+import com.databricks.labs.overwatch.ApiCallV2.sc
 import com.databricks.labs.overwatch.pipeline.PipelineFunctions
 import com.databricks.labs.overwatch.utils._
 import com.fasterxml.jackson.databind.ObjectMapper
@@ -12,7 +13,12 @@ import org.json.JSONObject
 import scalaj.http.{HttpOptions, HttpResponse}
 
 import java.util
+import java.util.Collections
+import java.util.concurrent.Executors
 import scala.annotation.tailrec
+import scala.concurrent.{ExecutionContext, ExecutionContextExecutor, Future}
+import scala.math.Ordered.orderingToOrdered
+import scala.util.{Failure, Success}
 
 /**
  * Companion object for APICallV2.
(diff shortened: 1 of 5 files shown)


src/main/scala/com/databricks/labs/overwatch/ApiCallV2.scala
```diff
@@ -640,6 +652,123 @@ class ApiCallV2(apiEnv: ApiEnv) extends SparkSessionWrapper {
...
+  /**
+   * Function to make parallel API calls. Currently this functions supports only SqlQueryHistory and ClusterEvents
+   * @param endpoint
+   * @param jsonInput
+   * @param config
+   * @return
+   */
+  def makeParallelApiCalls(endpoint: String, jsonInput: Map[String, String], config: Config): String = {
+    var startValue = jsonInput.get("start_value").get.toLong
+    val endValue = jsonInput.get("end_value").get.toLong
```
sriram251-code, line 664: all the jsonInput may not contain start_value and end_value. This piece of code can be more generic. " jsonInput.get("start_value").get.toLong" this can cause null pointer exception.

src/main/scala/com/databricks/labs/overwatch/ApiCallV2.scala
```diff
@@ -640,6 +652,123 @@ class ApiCallV2(apiEnv: ApiEnv) extends SparkSessionWrapper {
...
+    val incrementCounter = jsonInput.get("increment_counter").get.toLong
+    val finalResponseCount = jsonInput.get("final_response_count").get.toLong
```
sriram251-code, line 666: same goes to 
 val incrementCounter = jsonInput.get("increment_counter").get.toLong
 val finalResponseCount = jsonInput.get("final_response_count").get.toLong 
this can cause null pointer exception.

src/main/scala/com/databricks/labs/overwatch/ApiCallV2.scala
sriram251-code: i guess we should move the part specific to the API to APIMETA.

sriram251-code (changes requested):

src/main/scala/com/databricks/labs/overwatch/pipeline/Bronze.scala
```diff
@@ -139,7 +139,8 @@ class Bronze(_workspace: Workspace, _database: Database, _config: Config)
         config.organizationId,
         database,
         BronzeTargets.clusterEventsErrorsTarget,
-        config.tempWorkingDir
+        config.tempWorkingDir,
+        config
```
sriram251-code, line 143: remove  config.tempWorkingDir as argument as we are already passing config

sriram251-code (changes requested):

sriram251-code (approved):

GeekSheikh (approved):

PR #759, opened 2023-02-17
improved first run Impute clusterSpec

### LINKED ISSUE #758: [BUG] - First Run - ClusterSpec - Missing Some Record Initialization

GeekSheikh:
closes #758


brij-raghuwanshi-db (approved):
Looks good.
However, didn't @souravbaner-da implement these rankings somewhere recently? Did we miss it here or this is a completely different set/kind of issue and its relevant solution/change?

GeekSheikh:
@brij-raghuwanshi-db -- if you can work with @souravbaner-da to determine which PR he did this on we can discuss where the miss was. If you compare the old and new code you'll notice that several cluster ids were not getting identified for imputation on first run resulting in cost-critical components missing such as node type and runtime engine as they were not getting imputed. This resulted in additional price misses.

PR #773, opened 2023-02-20
spark conf handler plus AQE optimizations

### LINKED ISSUE #771: Par Spark Conf - Replicate global configs to thread-specific new sessions

GeekSheikh:
closes #771

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/env/Database.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/env/Database.scala
@@ -155,9 +155,9 @@ class Database(config: Config) extends SparkSessionWrapper {
          |""".stripMargin)
     df.write.format("delta").save(dfTempPath)
 
-    spark.conf.set("spark.sql.files.maxPartitionBytes", 1024 * 1024 * 16) // maximize parallelism on re-read and let
+    // maximize parallelism on re-read and let AQE bring it back down
+    spark.conf.set("spark.sql.files.maxPartitionBytes", 1024 * 1024 * 16)
     spark.conf.set("spark.databricks.delta.formatCheck.enabled", "true")
-    // AQE bring it back down
     spark.read.format("delta").load(dfTempPath)
   }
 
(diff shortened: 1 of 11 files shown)


sriram251-code:
Screenshot 2023-02-21 at 11 54 32 AM
I guess for generateBaseConfig also we need to set spark(glabalSession=true)

src/main/scala/com/databricks/labs/overwatch/env/Database.scala
```diff
@@ -155,9 +155,9 @@ class Database(config: Config) extends SparkSessionWrapper {
          |""".stripMargin)
     df.write.format("delta").save(dfTempPath)
 
-    spark.conf.set("spark.sql.files.maxPartitionBytes", 1024 * 1024 * 16) // maximize parallelism on re-read and let
+    // maximize parallelism on re-read and let AQE bring it back down
+    spark.conf.set("spark.sql.files.maxPartitionBytes", 1024 * 1024 * 16)
```
sriram251-code, line 159: why 16 mb ?

sriram251-code (changes requested):

GeekSheikh:
> Screenshot 2023-02-21 at 11 54 32 AM I guess for generateBaseConfig also we need to set spark(glabalSession=true)

doesn't this happen before `parSessionOn` is enabled thus it should just use global by default, right?

src/main/scala/com/databricks/labs/overwatch/env/Database.scala
GeekSheikh: I didn't change this. This is to increase parallelism on the persistAndLoad functions to get more parallelism during load and allow AQE to coalesce the partitions during the subsequent shuffles.

sriram251-code (approved):

PR #777, opened 2023-02-21
mount mapping validation added

### LINKED ISSUE #775: mount_mapping_path validation

sriram251-code:
closes #775

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/validation/DeploymentValidation.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/validation/DeploymentValidation.scala
@@ -98,23 +98,29 @@ object DeploymentValidation extends SparkSessionWrapper {
     }
   }
 
-  private def validateMountCount(conf: MultiWorkspaceConfig): DeploymentValidationReport = {
+  private def validateMountMappingPath(conf: MultiWorkspaceConfig): DeploymentValidationReport = {
 
-    val isAzure = conf.cloud.toLowerCase == "azure" //Mount-point validation is only done for Azure
-    val isRemoteWorkspace = conf.workspace_id.trim != Initializer.getOrgId // No need to perform mount-point validation for driver workspace.
+    val path = conf.mount_mapping_path.trim
+    val testDetails =
+      s"""WorkSpaceMountTest
+         |mount_mapping_path:${path}
+         """.stripMargin
+    try {
+      if (!Helpers.pathExists(path)) {
+        DeploymentValidationReport(false,
+          getSimpleMsg("Validate_Mount"),
+          testDetails,
+          Some("Unable to find the provided csv: " + path),
+          Some(conf.workspace_id)
+        )
+      } else {
+        val inputDf = spark.read.option("header", "true")
+          .option("ignoreLeadingWhiteSpace", true)
+          .option("ignoreTrailingWhiteSpace", true)
+          .csv(path)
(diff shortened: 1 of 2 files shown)


src/main/scala/com/databricks/labs/overwatch/validation/DeploymentValidation.scala
```diff
@@ -98,23 +98,29 @@ object DeploymentValidation extends SparkSessionWrapper {
...
+          testDetails,
+          Some("Unable to find the provided csv: " + path),
+          Some(conf.workspace_id)
+        )
+      } else {
+        val inputDf = spark.read.option("header", "true")
+          .option("ignoreLeadingWhiteSpace", true)
+          .option("ignoreTrailingWhiteSpace", true)
+          .csv(path)
+          .select("mountPoint", "source")
```
GeekSheikh, line 121: shouldn't we filter where organization_id == organization_id in the context?

```scala
.filter('workspaceId === conf.workspace_id)
```

src/main/scala/com/databricks/labs/overwatch/validation/DeploymentValidation.scala
```diff
@@ -98,23 +98,29 @@ object DeploymentValidation extends SparkSessionWrapper {
...
-           |APIURL:${conf.api_url}
-           |DBPATWorkspaceScope:${conf.secret_scope}
-           |SecretKey_DBPAT:${conf.secret_key_dbpat}""".stripMargin
-      try {
-        val patToken = dbutils.secrets.get(scope = conf.secret_scope, key = conf.secret_key_dbpat)
-        val apiEnv = ApiEnv(false, conf.api_url, patToken, getClass.getPackage.getImplementationVersion)
-        val endPoint = "dbfs/search-mounts"
-        val mountCount = ApiCallV2(apiEnv, endPoint).execute().asDF().count()
-        if (mountCount < 50) {
+        if (inputDf.count() > 0) {
```
GeekSheikh, line 123: let's put this count into a variable and log the number of paths found for the workspace in the logs as well as add the count to the message after `Some("SUCCESS")` thoughts?

GeekSheikh (changes requested):

src/main/scala/com/databricks/labs/overwatch/validation/DeploymentValidation.scala
sriram251-code: here we dont have workspaceId in mount_mapping_path csv file.

src/main/scala/com/databricks/labs/overwatch/validation/DeploymentValidation.scala
sriram251-code: the count is the number of mount to source mapping that has been found in the provided csv. we can add that but dont give us major help.

GeekSheikh: it does help because users may know approximately how many mounts they have by workspace so if they see validation Success and it says they have 1200 mounts but they know that workspace shouldn't have that many then they know they mapped the wrong file to the workspace, right?

GeekSheikh (approved):

PR #786, opened 2023-03-01
gcp integration added

### LINKED ISSUE #61: [EPIC] - Enable support for GCP
As we get closer to the GCP release, we'll need to consider adding the support for GCP (first step would be just to throw an error during initialization, saying that GCP is not supported)

sriram251-code:
closes #61 

Below is the test notebook
https://416411475796958.8.gcp.databricks.com/?o=416411475796958#notebook/1998104671077521/command/1998104671077524


src/main/scala/com/databricks/labs/overwatch/utils/Config.scala
```diff
@@ -303,7 +302,7 @@ class Config() {
     this
   }
 
-  private def setApiEnv(value: ApiEnv): this.type = {
+  def setApiEnv(value: ApiEnv): this.type = {
```
GeekSheikh: This needs to stay private

src/main/scala/com/databricks/labs/overwatch/utils/Config.scala
```diff
@@ -388,8 +397,73 @@ class Config() {
...
 
+  /**
+   * Function derives the cloud from workspaceUrl.
+   * @return
+   */
+  private def deriveCloudProvider(): String = {
+    if (_workspaceUrl.toLowerCase().contains("azure")) "azure"
+    else if (_workspaceUrl.toLowerCase().contains("aws")) "aws"
+    else "gcp"
+  }
```
GeekSheikh: else should not be "gcp". I added a commit to switch this to a case statement and fail if something other than supported cloud is identified. I think my new commit will break PVC though -- we need to look more closely at the isPVC rule and make sure we handle this -- all PVC will be AWS at this time.

src/main/scala/com/databricks/labs/overwatch/validation/DeploymentValidation.scala
```diff
@@ -378,6 +378,13 @@ object DeploymentValidation extends SparkSessionWrapper {
           config.eh_scope_key,
           config.eh_name,
           config.output_path)
+      case "gcp" =>
+        validateAuditLog(
+          config.workspace_id,
+          config.auditlogprefix_source_aws,
+          config.primordial_date,
+          config.max_days
+        )
```
GeekSheikh: added commit -- to avoid code redundancy merged logic into aws || gcp

src/main/scala/com/databricks/labs/overwatch/validation/DeploymentValidation.scala
```diff
@@ -378,6 +378,13 @@ object DeploymentValidation extends SparkSessionWrapper {
           config.eh_scope_key,
           config.eh_name,
           config.output_path)
+      case "gcp" =>
+        validateAuditLog(
+          config.workspace_id,
+          config.auditlogprefix_source_aws,
```
GeekSheikh: This config needs to be changed globally to rename from `auditlogprefix_source_aws` to `auditlogprefix_source_path`. This will need to be handled in the DOCS and in the upgrade script. Please add two tickets for 0720 with the appropriate labels to make sure these get done and change the config name in this PR.

GeekSheikh (changes requested):

src/main/scala/com/databricks/labs/overwatch/utils/Config.scala
sriram251-code: done

src/main/scala/com/databricks/labs/overwatch/utils/Config.scala
sriram251-code: changed the code to handle PVC

src/main/scala/com/databricks/labs/overwatch/validation/DeploymentValidation.scala
sriram251-code: Ticket created and assigned to 7.2.0
https://github.com/databrickslabs/overwatch/issues/802

src/main/scala/com/databricks/labs/overwatch/validation/DeploymentValidation.scala
sriram251-code: duplicate code removed

GeekSheikh (approved):
good work.

PR #795, opened 2023-03-02
disable autoOptimizeShuffle

### LINKED ISSUE #794: [PERF] - autoOptimizeShuffle was enabled and is causing perf issues
In an attempt to increase performance -- a regression was made that results in the larger modules (i.e. spark table merges) to have WAY too few tasks resulting in extremely long -- never finishing runtimes.

Solution -- disable this flag.

https://github.com/databrickslabs/overwatch/blob/ae56406ea5aef1e1e8a40b3653a9e6df70922458/src/main/scala/com/databricks/labs/overwatch/utils/SparkSessionWrapper.scala#L20

GeekSheikh:
closes #794


brij-raghuwanshi-db (approved):
Based on the discussion and scenario, it makes sense.

sriram251-code (approved):

PR #796, opened 2023-03-03
Enable API Error Reporting to PipReport and Safe Failure

### LINKED ISSUE #789: [BUG] - APICallV2 - Doesn't Handle Permission Denied
APICallV2 is killing the pipeline when non-admin token is used and the accounts scope is enabled.

tokenSnapshotModule and globalInitScSnapshotModule require admin but instead of failing the module, it's breaking the pipeline. Below is error and attached is full stack trace.

```
Exception: com.databricks.labs.overwatch.utils.ApiCallFailure: {"error_code":"PERMISSION_DENIED","message":"This API is reserved to admin users."}
Caused by: ApiCallFailure: {"error_code":"PERMISSION_DENIED","message":"This API is reserved to admin users."}
at com.databricks.labs.overwatch.ApiCallV2.execute(ApiCallV2.scala:664)
```
[stackTrace (2).txt](https://github.com/databrickslabs/overwatch/files/10866295/stackTrace.2.txt)

GeekSheikh:
closes #789 

Solution -- send ETLProcess into Module as a function inside the try block

Proven working

image


sriram251-code (approved):
awesome

PR #799, opened 2023-03-03
fixed exists tests

### LINKED ISSUE #787: [BUG] - PipelineTable.Exists -- logic error
when pathValidation == true and catalogValidation == true the logic can fail.

This is critical path because it breaks Database.registerTarget as the line below always returns false when target data exists but catalog table does not. This results in missing catalog tables never getting registered.

```scala 
if (!target.exists(catalogValidation = true) && target.exists(pathValidation = true))
```

GeekSheikh:
closes #787


souravbaner-da (approved):

brij-raghuwanshi-db (commented):
Looks good.

PR #801, opened 2023-03-06
622 verbose audit logging

souravbaner-da:
Add Verbose Audit Log

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/pipeline/Gold.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/pipeline/Gold.scala
@@ -3,6 +3,17 @@ package com.databricks.labs.overwatch.pipeline
 import com.databricks.labs.overwatch.env.{Database, Workspace}
 import com.databricks.labs.overwatch.utils.{Config, OverwatchScope}
 import org.apache.log4j.Logger
+import org.apache.spark.sql.functions.{col, from_unixtime}
+import com.databricks.labs.overwatch.pipeline.PipelineFunctions.spark.table
+import com.databricks.labs.overwatch.pipeline.TransformFunctions._
+import com.databricks.labs.overwatch.pipeline.WorkflowsTransforms._
+import com.databricks.labs.overwatch.utils.{NoNewDataException, SchemaTools, SparkSessionWrapper, TimeTypes}
+import org.apache.log4j.Level
+import org.apache.spark.sql.expressions.Window
+import org.apache.spark.sql.functions._
+import org.apache.spark.sql.{Column, DataFrame}
+import com.databricks.labs.overwatch.utils.{Config, SparkSessionWrapper}
+import org.apache.log4j.{Level, Logger}
 
 
 class Gold(_workspace: Workspace, _database: Database, _config: Config)
@@ -362,6 +373,53 @@ class Gold(_workspace: Workspace, _database: Database, _config: Config)
       case _ =>
     }
   }
+  private def publishVerbose():Unit = {
+    val notebookLookupTSDF = spark.sql(s"SELECT * FROM ${config.consumerDatabaseName}.notebook")
+      .select("organization_id", "notebook_id", "notebook_path", "notebook_name", "unixTimeMS")
+      .withColumnRenamed("notebook_id","notebookId")
+      .distinct
+      .toTSDF("unixTimeMS", "organization_id", "notebookId")
+
+    val clsfLookupTSDF = spark.sql(s"SELECT * FROM ${config.consumerDatabaseName}.clusterstatefact")
+      .select(
+        "organization_id", "state_start_date", "unixTimeMS_state_start", "cluster_id", "cluster_name",
+        "custom_tags", "node_type_id", "current_num_workers")
+      .distinct
+      .withColumnRenamed("state_start_date","date")
+      .withColumnRenamed("unixTimeMS_state_start","unixTimeMS")
+      .withColumnRenamed("cluster_id","clusterId")
+      .withColumnRenamed("current_num_workers","node_count")
+      .withColumn("totalCostPMS",col("uptime_in_state_H")/ col("total_cost")/lit(60)/lit(60)/lit(1000))
+      .toTSDF("unixTimeMS", "organization_id", "date", "clusterId")
+
+    val notebookCmdColNames = Array(
+      "organization_id", "workspace_name", "date", "timestamp", "userIdentity.email",
+      "notebookId", "notebook_path", "notebook_name", "commandId", "commandText", "executionTime","sourceIpAddress",
+      "userIdentity","shardName","execution_estimated_cost", "status", "clusterId", "cluster_name", "custom_tags",
+      "node_type_id", "node_count","sourceIPAddress", "response", "userAgent", "unixTimeMS"
+    )
+
+    val notebookCodeAndMetaDF = spark.sql(s"SELECT * FROM ${config.databaseName}.audit_log_bronze")
+      //   .filter('organization_id.isin(orgIDs: _*) && 'serviceName === "notebook" && 'actionName === "runCommand")
+      .filter((col("serviceName") === "databrickssql" && col("actionName") === "commandSubmit")||
+        (col("serviceName") === "databrickssql" && col("actionName") === "commandFinish") ||
+        (col("serviceName") === "notebook" && col("actionName") === "runCommand"))
+      .verifyMinimumSchema(Schema.auditMasterSchema)
+      .selectExpr("*", "requestParams.*").drop("requestParams")
+      .withColumnRenamed("timestamp", "unixTimeMS")
+      .withColumn("timestamp", from_unixtime(col("unixTimeMS") / 1000.0).cast("timestamp"))
+      .toTSDF("unixTimeMS", "organization_id", "notebookId")
+      .lookupWhen(notebookLookupTSDF)
+      .df
+      .toTSDF("unixTimeMS", "organization_id", "date", "clusterId")
+      .lookupWhen(clsfLookupTSDF)
+      .df
+      .withColumn("execution_estimated_cost", col("executionTime") * col("totalCostPMS"))
+      .select(notebookCmdColNames map col: _*)
+    println("notebookCodeAndMetaDF is")
+    println(notebookCodeAndMetaDF.show(20,false))
+
+  }
 
   def refreshViews(workspacesAllowed: Array[String] = Array()): Unit = {
     config.overwatchScope.foreach {
@@ -407,6 +465,7 @@ class Gold(_workspace: Workspace, _database: Database, _config: Config)
     restoreSparkConf()
     executeModules()
     buildFacts()
+    publishVerbose()
 
     initiatePostProcessing()
     this // to be used as fail switch later if necessary
--- a/src/main/scala/com/databricks/labs/overwatch/pipeline/Gold.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/pipeline/Gold.scala
@@ -374,51 +374,59 @@ class Gold(_workspace: Workspace, _database: Database, _config: Config)
     }
   }
   private def publishVerbose():Unit = {
-    val notebookLookupTSDF = spark.sql(s"SELECT * FROM ${config.consumerDatabaseName}.notebook")
-      .select("organization_id", "notebook_id", "notebook_path", "notebook_name", "unixTimeMS")
-      .withColumnRenamed("notebook_id","notebookId")
-      .distinct
-      .toTSDF("unixTimeMS", "organization_id", "notebookId")
+    if((spark.catalog.tableExists(s"${config.consumerDatabaseName}.notebook"))
+      & (spark.catalog.tableExists("overwatch_dev_5481.notebook"))) {
+      val notebookLookupTSDF = spark.sql(s"SELECT * FROM ${config.consumerDatabaseName}.notebook")
+        .select("organization_id", "notebook_id", "notebook_path", "notebook_name", "unixTimeMS")
+        .withColumnRenamed("notebook_id", "notebookId")
+        .distinct
+        .toTSDF("unixTimeMS", "organization_id", "notebookId")
 
-    val clsfLookupTSDF = spark.sql(s"SELECT * FROM ${config.consumerDatabaseName}.clusterstatefact")
-      .select(
-        "organization_id", "state_start_date", "unixTimeMS_state_start", "cluster_id", "cluster_name",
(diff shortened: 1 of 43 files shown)


src/main/scala/com/databricks/labs/overwatch/pipeline/Gold.scala
```diff
@@ -362,6 +373,62 @@ class Gold(_workspace: Workspace, _database: Database, _config: Config)
...
+        .toTSDF("unixTimeMS", "organization_id", "notebookId")
+        .lookupWhen(notebookLookupTSDF)
+        .df
+        .toTSDF("unixTimeMS", "organization_id", "date", "clusterId")
+        .lookupWhen(clsfLookupTSDF)
+        .df
+        .withColumn("execution_estimated_cost", col("executionTime") * col("totalCostPMS"))
+        .select(ColNames map col: _*)
+
+      notebookCodeAndMetaDF.write.format("delta").mode("overwrite").save(s"${config.etlDataPathPrefix}/verbose_audit_log")
```
GeekSheikh, line 422: If you're going to materialize this -- you need to create a target and do incrementals -- doing a full compute and overwrite each time will crush perf in most customers. If you don't want / need to materialize this data you can just create a view.

src/main/scala/com/databricks/labs/overwatch/pipeline/Gold.scala
```diff
@@ -407,6 +474,7 @@ class Gold(_workspace: Workspace, _database: Database, _config: Config)
     restoreSparkConf()
     executeModules()
     buildFacts()
+    publishVerbose()
```
GeekSheikh, line 477: There's no try/catch here at all -- so if something were to fail in this function it would break the pipeline.

src/main/scala/com/databricks/labs/overwatch/pipeline/Gold.scala
```diff
@@ -362,6 +373,62 @@ class Gold(_workspace: Workspace, _database: Database, _config: Config)
       case _ =>
     }
   }
+  private def publishVerbose():Unit = {
+    if((spark.catalog.tableExists(s"${config.consumerDatabaseName}.notebook"))
+      & (spark.catalog.tableExists("overwatch_dev_5481.notebook"))) {
```
GeekSheikh, line 378: you have direct references to database name here.

GeekSheikh (changes requested):
ultimately this needs to either be a materialized target with appropriate incremental loads OR it needs to be a view -- we can't just throw a function at the end to build stuff. This won't be maintainable and doesn't fit in with the rest of the pipeline.

src/main/scala/com/databricks/labs/overwatch/pipeline/Gold.scala
```diff
@@ -363,6 +381,71 @@ class Gold(_workspace: Workspace, _database: Database, _config: Config)
...
+          "custom_tags", "node_type_id", "current_num_workers", "uptime_in_state_H", "total_cost")
+        .distinct
+        .withColumnRenamed("state_start_date", "date")
+        .withColumnRenamed("unixTimeMS_state_start", "unixTimeMS")
+        .withColumnRenamed("cluster_id", "clusterId")
+        .withColumnRenamed("current_num_workers", "node_count")
+        .withColumn("totalCostPMS", col("uptime_in_state_H") / col("total_cost") / lit(60) / lit(60) / lit(1000))
+        .toTSDF("unixTimeMS", "organization_id", "date", "clusterId")
+
+      val ColNames = Array(
```
GeekSheikh, line 410: any metric field name like execution_time should have UOM (unit of Measure) in this case it is _s

executionTime --> execution_time_S
execution_estimated_cost --> estimated_dbu_cost

src/main/scala/com/databricks/labs/overwatch/pipeline/Gold.scala
```diff
@@ -363,6 +381,71 @@ class Gold(_workspace: Workspace, _database: Database, _config: Config)
...
+        .toTSDF("unixTimeMS", "organization_id", "notebookId")
+        .lookupWhen(notebookLookupTSDF)
+        .df
+        .toTSDF("unixTimeMS", "organization_id", "date", "clusterId")
+        .lookupWhen(clsfLookupTSDF)
+        .df
+        .withColumn("execution_estimated_cost", col("executionTime") * col("totalCostPMS"))
+        .select(ColNames map col: _*)
+
+      notebookCodeAndMetaDF.write.format("delta").mode("overwrite").save(s"${config.etlDataPathPrefix}/verbose_audit_log")
```
GeekSheikh, line 435: See my earlier comment -- please follow the pipeline pattern

src/main/scala/com/databricks/labs/overwatch/pipeline/Gold.scala
```diff
@@ -363,6 +381,71 @@ class Gold(_workspace: Workspace, _database: Database, _config: Config)
...
+      spark.sql(s"create or replace view ${config.consumerDatabaseName}.verbose_audit_log as select * from delta.`${config.etlDataPathPrefix}/verbose_audit_log`")
```
GeekSheikh, line 436: you need to follow the pipeline pattern here too -- you're missing partition filter push down by doing it this way -- something that is natively handled for you if you just add this to the standard view publication pattern.

src/main/scala/com/databricks/labs/overwatch/pipeline/Gold.scala
```diff
@@ -407,6 +490,7 @@ class Gold(_workspace: Workspace, _database: Database, _config: Config)
     restoreSparkConf()
     executeModules()
     buildFacts()
+    verboseAudit()
```
GeekSheikh, line 493: This should not be here -- when you follow the pattern this logic will become part of refreshViews. 

Just add case to the very end and make a complex case statement
case audit + notebooks + clusterEvents --> publish this view

...
case audit & notebooks & clusterEvents => {GoldTargets.viewTarget.publish(...)}

The above is obviously pseudocode -- you may have to do a logical operator func and use if for this case

src/main/scala/com/databricks/labs/overwatch/pipeline/Gold.scala
```diff
@@ -110,6 +113,21 @@ class Gold(_workspace: Workspace, _database: Database, _config: Config)
...
   }
 
+  lazy private[overwatch] val verboseAuditModule = Module(3018, "Gold_VerboseAuditLog", this, Array(1004,3004,3005))
+  lazy private val appendVerboseAuditProccess: () => ETLDefinition = {
+    () =>
+      ETLDefinition(
+        BronzeTargets.auditLogsTarget.asIncrementalDF(verboseAuditModule, BronzeTargets.auditLogsTarget.incrementalColumns),
+        Seq(buildVerboseAudit(
+          GoldTargets.notebookTarget,
+          GoldTargets.clusterStateFactTarget
```
GeekSheikh, line 123: this must be incremental DFs

CLSF should have lag of 30d to ensure long-running states are within scope but not forever history

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
```diff
@@ -237,6 +237,59 @@ trait GoldTransforms extends SparkSessionWrapper {
       )
   }
 
+  protected def buildVerboseAudit(
```
GeekSheikh, line 240: This looks a lot better than all of the logic being in gold -- not sure why the logic is all duplicated -- Please take the requested changes from Gold.scala and implement there here

src/main/scala/com/databricks/labs/overwatch/pipeline/Gold.scala
```diff
@@ -363,6 +381,71 @@ class Gold(_workspace: Workspace, _database: Database, _config: Config)
...
+      val clsfLookupTSDF = spark.sql(s"SELECT * FROM ${config.consumerDatabaseName}.clusterstatefact")
+        .select(
+          "organization_id", "state_start_date", "unixTimeMS_state_start", "cluster_id", "cluster_name",
+          "custom_tags", "node_type_id", "current_num_workers", "uptime_in_state_H", "total_cost")
+        .distinct
+        .withColumnRenamed("state_start_date", "date")
+        .withColumnRenamed("unixTimeMS_state_start", "unixTimeMS")
+        .withColumnRenamed("cluster_id", "clusterId")
+        .withColumnRenamed("current_num_workers", "node_count")
+        .withColumn("totalCostPMS", col("uptime_in_state_H") / col("total_cost") / lit(60) / lit(60) / lit(1000))
```
GeekSheikh, line 407: I believe this logic is off as the final cost metric is showing very silly numbers.

src/main/scala/com/databricks/labs/overwatch/pipeline/Gold.scala
GeekSheikh, line 410: don't start a variable with a capital letter

GeekSheikh, line 410: what is shardName and why is it here? It seems to always be null and shouldn't it basically be the same thing as workspace_name (semantically)? Suggest remove this field and only use workspace_name

src/main/scala/com/databricks/labs/overwatch/pipeline/PipelineTargets.scala
```diff
@@ -498,6 +498,22 @@ abstract class PipelineTargets(config: Config) {
       config
     )
 
+    lazy private[overwatch] val verboseAuditTarget: PipelineTable = PipelineTable(
```
GeekSheikh, line 501: if you continue to use MERGE -- _permitDuplicateKeys should be false unless I'm missing something

src/main/scala/com/databricks/labs/overwatch/pipeline/PipelineTargets.scala
```diff
@@ -498,6 +498,22 @@ abstract class PipelineTargets(config: Config) {
...
+      name = "verboseAuditTarget_gold",
+      _keys = Array("organization_id", "timestamp", "unixTimeMS"),
+      config,
+      _mode = WriteMode.merge,
+      partitionBy = Seq("organization_id"),
+      incrementalColumns = Array("unixTimeMS"),
+      zOrderBy = Array("notebookId", "unixTimeMS")
```
GeekSheikh, line 508: zorders are expensive and I don't really see many people (or downstream ETLs) referencing notebookId or unixTimeMS heavily, do you? Zorders are meant to speed up ETL and/or end-user queries -- I foresee people filtering on notebookPath and using wildcards and/or the actual timestamp column much more than unixTimeMS. 

If you agree -- I suggest removing these orders -- if you disagree, please make your case and we can review.

src/main/scala/com/databricks/labs/overwatch/pipeline/PipelineTargets.scala
```diff
@@ -498,6 +498,22 @@ abstract class PipelineTargets(config: Config) {
       config
     )
 
+    lazy private[overwatch] val verboseAuditTarget: PipelineTable = PipelineTable(
+      name = "verboseAuditTarget_gold",
+      _keys = Array("organization_id", "timestamp", "unixTimeMS"),
```
GeekSheikh, line 503: why are timestamp AND unixTimeMS BOTH keys? unixTimeMS is the same as timestamp but with slightly higher resolution -- as such remove timestamp from keys.

src/main/scala/com/databricks/labs/overwatch/pipeline/PipelineTargets.scala
```diff
@@ -498,6 +498,22 @@ abstract class PipelineTargets(config: Config) {
...
+      config,
+      _mode = WriteMode.merge,
```
GeekSheikh, line 505: Why MERGE? Merge is very expensive and the actions are immutable right? By removing dbsql -- you can get rid of potentially getting a start in overwatch run 1 and a finish in run 2 and greatly simplify this. All of the DBSQL query information is captured in SQL_query_history table (and more); thus it doesn't need to be captured here.

src/main/scala/com/databricks/labs/overwatch/pipeline/PipelineTargets.scala
```diff
@@ -498,6 +498,22 @@ abstract class PipelineTargets(config: Config) {
       config
     )
 
+    lazy private[overwatch] val verboseAuditTarget: PipelineTable = PipelineTable(
+      name = "verboseAuditTarget_gold",
```
GeekSheikh, line 502: change this name to something friendlier...perhaps notebookCommands_gold (feel free to find something better)

certainly we don't want to put the word "audit" in a gold table -- customers may have misplaced expectations

src/main/scala/com/databricks/labs/overwatch/pipeline/PipelineTargets.scala
```diff
@@ -498,6 +498,22 @@ abstract class PipelineTargets(config: Config) {
...
+      _keys = Array("organization_id", "timestamp", "unixTimeMS"),
+      config,
+      _mode = WriteMode.merge,
+      partitionBy = Seq("organization_id"),
+      incrementalColumns = Array("unixTimeMS"),
+      zOrderBy = Array("notebookId", "unixTimeMS")
+    )
+
+    lazy private[overwatch] val verboseAuditViewTarget: PipelineView = PipelineView(
+      name = "verboseAuditTarget",
```
GeekSheikh, line 512: change name to be appropriate for whatever you wind up naming the gold table, just remove _gold

GeekSheikh (changes requested):

src/main/scala/com/databricks/labs/overwatch/pipeline/Gold.scala
```diff
@@ -363,6 +381,71 @@ class Gold(_workspace: Workspace, _database: Database, _config: Config)
...
+        .toTSDF("unixTimeMS", "organization_id", "date", "clusterId")
+
+      val ColNames = Array(
+        "organization_id", "workspace_name", "date", "timestamp", "userIdentity.email",
+        "notebookId", "notebook_path", "notebook_name", "commandId", "commandText", "executionTime", "sourceIpAddress",
+        "userIdentity", "shardName", "execution_estimated_cost", "status", "clusterId", "cluster_name", "custom_tags",
+        "node_type_id", "node_count", "response", "userAgent", "unixTimeMS"
+      )
+
+      val notebookCodeAndMetaDF = spark.sql(s"SELECT * FROM delta.`${config.etlDataPathPrefix}/audit_log_bronze`")
```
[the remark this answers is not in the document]
souravbaner-da: Do we need to expose custom_tags here in this table

souravbaner-da:
Duplicate PR already there #917

PR #805, opened 2023-03-08
Enable Incremental in Bronze Snapshot

souravbaner-da:
Enable Incremental in Bronze Snapshot


src/main/scala/com/databricks/labs/overwatch/pipeline/Bronze.scala
```diff
@@ -95,6 +95,37 @@ class Bronze(_workspace: Workspace, _database: Database, _config: Config)
     Helpers.parClone(cloneSpecs)
 
   }
+  def snapshotStream(
```
GeekSheikh, line 98: I think this needs to be moved to workspace and called `incrementalSnap`

currently users backup using `workspace.snap` after this is merged customers should be able to call `workspace.incrementalSnap`

GeekSheikh, line 98: Add code docs on what is expected to be passed in for these variables and an English explanation of what this function does.

GeekSheikh, line 98: be sure to confirm that the snapshot backs up ALL data from all organization_ids. This means you need to test on a workspace that has multiple workspaces in the same bronze tables.

src/main/scala/com/databricks/labs/overwatch/pipeline/Bronze.scala
```diff
@@ -95,6 +95,37 @@ class Bronze(_workspace: Workspace, _database: Database, _config: Config)
...
+                      targetPrefix: String,
```
GeekSheikh, line 99: I think we should change the name of this in this func and the existing .snap func. It's too close to target prefix meaning PipelineTable ETLPrefix. I think this should be renamed to something like `snapshotRootPath` or something to make it more clear.

src/main/scala/com/databricks/labs/overwatch/pipeline/Bronze.scala
```diff
@@ -95,6 +95,37 @@ class Bronze(_workspace: Workspace, _database: Database, _config: Config)
...
+      if (exclude.contains(".")) exclude.split("\\.").takeRight(1).head else exclude
+    })
+
+    // remove excludes
+    // remove non-existing bronze targets
+    val targetsToSnap = bronzeTargets
+      .filter(_.exists()) // source path must exist
+      .filterNot(t => cleanExcludes.contains(t.name.toLowerCase))
+
+    val finalTargetPathPrefix = s"${targetPrefix}/bckup"
```
GeekSheikh, line 115: why are we nesting this into a subdir here?

src/main/scala/com/databricks/labs/overwatch/utils/Tools.scala
```diff
@@ -436,6 +437,44 @@ object Helpers extends SparkSessionWrapper {
     spark.conf.set("spark.databricks.delta.vacuum.parallelDelete.enabled", "false")
     s"SHRED COMPLETE: ${target.tableFullName}"
   }
+  /**
+   * Execute a parallelized streaming to follow the instructions provided through CloneDetail class
+   *
+   * @param cloneDetails details required to execute the parallelized clone
+   * @return
+   */
+  def tableStream(cloneDetails: Seq[CloneDetail]): Seq[CloneReport] = {
```
GeekSheikh, line 446: this should be private[overwatch] I think since I don't think we want to allow use outside.

let's rename the func to make it more clear that it's only a snapStream.

I'm on the fence as to whether we should have a class/object called snapshot and refactor the core logic out of this Helpers section. We could move snap and this new `snapStream` over to the other location. We could have a Helper function that then instantiates/calls the core logic from Snapshot.scala or something. What do you think?

src/main/scala/com/databricks/labs/overwatch/pipeline/Bronze.scala
GeekSheikh: I think this makes sense if it were called something like `s"${targetPrefix}/data"` which would allow for another folder called `s"${targetPrefix}/report"` where we could persist the snapshot reports.

src/main/scala/com/databricks/labs/overwatch/utils/Tools.scala
GeekSheikh, line 446: I think we need to persist the clone report. This brings more credence to the idea of refactoring the snapshots out of these helper sections to allow for a more complex / controlled workflow to happen. I even think there could be a main class for snapshotting since everyone will be doing this from a scheduled job anyway.

src/main/scala/com/databricks/labs/overwatch/pipeline/Bronze.scala
```diff
@@ -95,6 +95,37 @@ class Bronze(_workspace: Workspace, _database: Database, _config: Config)
...
+    // !Overwrite - targetPrefix/currDateString/timestampMillisString
+
+    // build clone details
+    val cloneSpecs = targetsToSnap.map(t => {
+      val targetPath = s"${finalTargetPathPrefix}/${t.name.toLowerCase}"
+      CloneDetail(t.tableLocation, targetPath)
+    })
```
GeekSheikh, line 122: this looks like reusable code -- move to a function?

GeekSheikh (changes requested):

src/main/scala/com/databricks/labs/overwatch/pipeline/Snapshot.scala
```diff
@@ -0,0 +1,115 @@
...
+    val cloneSpecs = targetsToSnap.map(dataset => {
+      val sourceName = dataset.name.toLowerCase
+      val sourcePath = dataset.tableLocation
+      val targetPath = if (snapshotRootPath.takeRight(1) == "/") s"$snapshotRootPath$sourceName" else s"$snapshotRootPath/$sourceName"
+      CloneDetail(sourcePath, targetPath, asOfTS, cloneLevel)
+    }).toArray.toSeq
+    cloneSpecs
+  }
+
+  def main(args: Array[String]): Unit = {
```
sriram251-code, line 97: add documentation of the function

src/main/scala/com/databricks/labs/overwatch/pipeline/Snapshot.scala
```diff
@@ -0,0 +1,115 @@
...
+    })
+    val targetsToSnap = bronzeTargets
+      .filter(_.exists()) // source path must exist
+      .filterNot(t => cleanExcludes.contains(t.name.toLowerCase))
+    val finalSnapshotRootPath = s"${snapshotRootPath}/data"
+    val cloneSpecs = buildCloneSpecs(targetsToSnap, finalSnapshotRootPath)
+    Helpers.snapStream(cloneSpecs, snapshotRootPath)
+  }
+
+  def buildCloneSpecs(
```
sriram251-code, line 81: make it private

sriram251-code (changes requested):

src/main/scala/com/databricks/labs/overwatch/pipeline/Snapshot.scala
```diff
@@ -0,0 +1,124 @@
...
+   *                         far as the latest vacuum by design.
+   * @param excludes         Array of table names to exclude from the snapshot
+   *                         this is the table name only - without the database prefix
+   * @return
+   */
+  private[overwatch] def snap(
+              bronze: Bronze,
+              workspace: Workspace,
+              snapshotRootPath: String,
+              cloneLevel: String = "DEEP",
```
GeekSheikh, line 30: This should be able to be overridden in the companion object apply method -- as of now, it's always default and unchangeable outside of Overwatch package.

src/main/scala/com/databricks/labs/overwatch/pipeline/Snapshot.scala
```diff
@@ -0,0 +1,124 @@
...
+   * @param cloneLevel       Deep or Shallow
+   * @param asOfTS           appends asOfTimestamp option to Delta reader to limit data on clone. This will only go back as
+   *                         far as the latest vacuum by design.
+   * @param excludes         Array of table names to exclude from the snapshot
+   *                         this is the table name only - without the database prefix
+   * @return
+   */
+  private[overwatch] def snap(
+              bronze: Bronze,
+              workspace: Workspace,
```
GeekSheikh, line 28: You're passing in workspace but never using it.

src/main/scala/com/databricks/labs/overwatch/pipeline/Snapshot.scala
```diff
@@ -0,0 +1,124 @@
...
+   *
+   * @param snapshotRootPath prefix of path target to send the snap
+   * @param excludes         Array of table names to exclude from the snapshot
+   *                         this is the table name only - without the database prefix
+   * @return
+   */
+
+  private[overwatch] def incrementalSnap(
+                       bronze: Bronze,
+                       workspace: Workspace,
```
GeekSheikh, line 63: You're passing in workspace but never using it.

src/main/scala/com/databricks/labs/overwatch/pipeline/Snapshot.scala
```diff
@@ -0,0 +1,124 @@
...
+      .filter(_.exists()) // source path must exist
+      .filterNot(t => cleanExcludes.contains(t.name.toLowerCase))
+    val finalSnapshotRootPath = s"${snapshotRootPath}/data"
+    val cloneSpecs = buildCloneSpecs(targetsToSnap, finalSnapshotRootPath)
+    Helpers.snapStream(cloneSpecs, snapshotRootPath)
+  }
+
+  private[overwatch] def buildCloneSpecs(
+                       targetsToSnap: Array[PipelineTable],
+//                       sourcesToSnap: Seq[WorkspaceDataset],
```
GeekSheikh, line 82: delete if not using.

src/main/scala/com/databricks/labs/overwatch/pipeline/Bronze.scala
```diff
@@ -35,6 +35,10 @@ class Bronze(_workspace: Workspace, _database: Database, _config: Config)
       BronzeTargets.jobRunsSnapshotTarget
     )
   }
+  def targetToSnap: Array[PipelineTable] = {
+    getAllTargets :+ pipelineStateTarget
+  }
+
```
GeekSheikh, line 41: don't put this here -- and definitely don't make it public. targetsToSnap should be a class variable inside of Snapshot class.

src/main/scala/com/databricks/labs/overwatch/pipeline/Bronze.scala
```diff
@@ -35,6 +35,10 @@ class Bronze(_workspace: Workspace, _database: Database, _config: Config)
       BronzeTargets.jobRunsSnapshotTarget
     )
   }
+  def targetToSnap: Array[PipelineTable] = {
```
GeekSheikh, line 38: Since we're moving all the logic from Bronze to a new class Snapshot we should drop all the logic from Bronze so we can stop maintaining it. For backwards compatibility simply reference the appropriate snapshot apply method to create a full (batch) snapshot so that it continues to work and add the @deprecated decorator.

src/main/scala/com/databricks/labs/overwatch/pipeline/Snapshot.scala
```diff
@@ -0,0 +1,124 @@
...
+   *                         this is the table name only - without the database prefix
+   * @return
+   */
+  private[overwatch] def snap(
+              bronze: Bronze,
+              workspace: Workspace,
+              snapshotRootPath: String,
+              cloneLevel: String = "DEEP",
+              asOfTS: Option[String] = None,
+              excludes: Array[String] = Array()
```
GeekSheikh, line 32: user has no way to control this from main class. Add arg for this

src/main/scala/com/databricks/labs/overwatch/pipeline/Snapshot.scala
```diff
@@ -0,0 +1,124 @@
...
+  private[overwatch] def incrementalSnap(
+                       bronze: Bronze,
+                       workspace: Workspace,
+                       snapshotRootPath: String,
+                       excludes: Array[String] = Array()
+                     ): Unit = {
+
+    val bronzeTargets = bronze.targetToSnap
+    val cleanExcludes = excludes.map(_.toLowerCase).map(exclude => {
+      if (exclude.contains(".")) exclude.split("\\.").takeRight(1).head else exclude
```
GeekSheikh, line 70: use the new path slash helpers introduced in #825 -- these are merged to 0720 now so rebase and you should have access to them.

src/main/scala/com/databricks/labs/overwatch/pipeline/Snapshot.scala
```diff
@@ -0,0 +1,124 @@
...
+    })
+    val targetsToSnap = bronzeTargets
+      .filter(_.exists()) // source path must exist
+      .filterNot(t => cleanExcludes.contains(t.name.toLowerCase))
+    val finalSnapshotRootPath = s"${snapshotRootPath}/data"
+    val cloneSpecs = buildCloneSpecs(targetsToSnap, finalSnapshotRootPath)
+    Helpers.snapStream(cloneSpecs, snapshotRootPath)
+  }
+
+  private[overwatch] def buildCloneSpecs(
```
GeekSheikh, line 80: Move this above its first reference in this file.

src/main/scala/com/databricks/labs/overwatch/utils/Tools.scala
```diff
@@ -437,13 +438,63 @@ object Helpers extends SparkSessionWrapper {
...
+          .trigger(Trigger.Once())
+          .option("checkpointLocation", checkPointLocation)
+          .option("path", s"${cloneSpec.target}")
+          .start()
+          .awaitTermination()
+        logger.log(Level.INFO, s"Streaming COMPLETE: ${cloneSpec.source} --> ${cloneSpec.target}")
+        CloneReport(cloneSpec, s"Streaming For: ${cloneSpec.source} --> ${cloneSpec.target}", "SUCCESS")
+      } catch {
+        case e: Throwable if (e.getMessage.contains("is after the latest commit timestamp of")) => {
+          val msg = s"SUCCESS WITH WARNINGS: The timestamp provided, ${cloneSpec.asOfTS.get} " +
```
GeekSheikh, line 476: Use the `appendStackStrace` method shown in other comment to build this message with the full stack trace as well

src/main/scala/com/databricks/labs/overwatch/utils/Tools.scala
```diff
@@ -437,13 +438,63 @@ object Helpers extends SparkSessionWrapper {
...
+    val taskSupport = new ForkJoinTaskSupport(new ForkJoinPool(parallelism))
+    cloneDetailsPar.tasksupport = taskSupport
+
+    logger.log(Level.INFO, "Streaming START:")
+    val cloneReport = cloneDetailsPar.map(cloneSpec => {
+      try {
+        val rawStreamingDF = spark.readStream.format("delta").load(s"${cloneSpec.source}")
+        val sourceName = s"${cloneSpec.source}".split("/").takeRight(1).head
+        val checkPointLocation = if (snapshotRootPath.takeRight(1) == "/") s"${snapshotRootPath}checkpoint/${sourceName}" else s"${snapshotRootPath}/checkpoint/${sourceName}"
+        val cloneReportPath = if (snapshotRootPath.takeRight(1) == "/") s"${snapshotRootPath}report/" else s"${snapshotRootPath}/clone_report/"
```
GeekSheikh, line 460: use the new path slash helpers introduced in https://github.com/databrickslabs/overwatch/pull/825 -- these are merged to 0720 now so rebase and you should have access to them.

GeekSheikh (changes requested):
Is there a proof for this? Don't just test that it runs -- please test that the data is incrementally appearing in the output, that it looks correct, and that the restore functionality works as expected.

souravbaner-da:
Duplicate PR already raised from Release-721
#909

PR #808, opened 2023-03-13
781 audit log format change

### LINKED ISSUE #781: [BUG] - MW - AWS - nonJSON audit log formats
When building the configs for aws AuditLogConfig the json format was not added. This config should be optional and should be exposed in the config.csv file. Default should be JSON.

Added quick fix in 711 release. The format can be changed by adding the following spark conf to cluster. Set it to `json`, `parquet`, etc. Only necessary when audit log delivery is not default (json).

`overwatch.aws.auditlogformat`

https://github.com/databrickslabs/overwatch/blob/afb1f9ebcaeae3351dd963d3a140b56234643719/src/main/scala/com/databricks/labs/overwatch/MultiWorkspaceDeployment.scala#L114-L117

sriram251-code:
closes #781


GeekSheikh:
@sriram251-code -- please file docs ticket to make sure we update the docs on how to do this.

PR #822, opened 2023-03-22
JRCP - Worker Potential H and Cluster Util Fixed

### LINKED ISSUE #821: [BUG] - JRCP - worker_potential and cluster_util - parallel tasks
When clusters have concurrency the workper_potential_core_h is not being calculated correctly. As a result, cluster_utilization can also be off.

CAUSE: the code is doing a simple lookup to clsf worker_potential_core_h and not discounting it for concurrency 
https://github.com/databrickslabs/overwatch/blob/cbf4cd2577e7a9b840ad9a9b6f96c07f4aa98346/src/main/scala/com/databricks/labs/overwatch/pipeline/WorkflowsTransforms.scala#L1518

GeekSheikh:
closes #821


aman-db (approved):
looks good to me

PR #826, opened 2023-03-23
changing code to derive cloud type

### LINKED ISSUE #820: Redefine the logic to derive cloud provider
As per the current logic cloud provider is getting derived using apiURL using the below logic - 
```
workspaceURL.toLowerCase match {
  case cloudType if cloudType.contains("azure") => "azure"
  case cloudType if cloudType.contains("aws") => "aws"
  case cloudType if cloudType.contains("gcp") => "gcp"
  case _ => throw new BadConfigException(s"${workspaceURL.toLowerCase} NOT SUPPORTED: Supported clouds at this time " +
    s"include azure, aws, gcp")
}
```
which is failing for aws apiURL's. Need to come up with a different logic to get cloud provider.

aman-db:
closes #820


GeekSheikh (approved):
Doing rapid approval to get 0720 fixed -- we will catch any issues with this during integration testing if this causes issues. But the logic makes sense to me.

PR #834, opened 2023-03-24
wrong dbu calculation for worker node for long running cluster

### LINKED ISSUE #803: [BUG] -Wrong dbu calculation for worker node for long running cluster.
Issue:

For long running cluster in CLSF the worker node is coming as 0 hence the worker_dbus are coming as 0.

Screenshot 2023-03-07 at 1 36 13 PM

This is happening because in cluster_events_bronze the num_workers are coming as null
Screenshot 2023-03-07 at 1 38 08 PM

Solution:
We can see num_workers perfectly populated in clusters_snapshot_bronze.
Screenshot 2023-03-07 at 1 40 03 PM
In 1st run for Overwatch if we find num_workers are coming as null in cluster_events_bronze then we should consider num_workers which is present in clusters_snapshot_bronze.

souravbaner-da:
Populate Number of Workers from Cluster_Snapshot_Bronze to CLSF.
Closes #803

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
@@ -334,6 +334,12 @@ trait GoldTransforms extends SparkSessionWrapper {
       // filling data is critical here to ensure jrcp dups don't occur
       // using noise generators to fill with two passes to reduce skew
       .fillMeta(clusterPotMetaToFill, clusterPotKeys, clusterPotIncrementals, noiseBuckets = getTotalCores) // scan back then forward to fill all
+      // Populate Num_workers from cluster_snap_bronze
+      .alias("clusterPotential")
+      .join(clusterSnapshot.asDF.alias("clusterSnapshot"),Seq("cluster_id"),"left").select("clusterPotential.*","clusterSnapshot.num_workers")
+      .withColumn("current_num_workers",when(col("current_num_workers") === "0",col("num_workers")).otherwise(col("current_num_workers")))
+      .withColumn("target_num_workers",when(col("target_num_workers") === "0",col("num_workers")).otherwise(col("target_num_workers")))
+      .drop("num_workers")
 
 
     val workerPotentialCoreS = when('databricks_billable, $"workerSpecs.vCPUs" * 'current_num_workers * 'uptime_in_state_S).otherwise(lit(0))
--- a/src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
@@ -279,7 +279,7 @@ trait GoldTransforms extends SparkSessionWrapper {
 
     val derivedSKU = PipelineFunctions.deriveSKU(isAutomated('cluster_name), 'spark_version, 'cluster_type)
     val nodeTypeLookup = clusterSpec.asDF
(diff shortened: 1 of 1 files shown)


GeekSheikh (changes requested):
The variable nodeTypeLookup2 is already doing a time-series lookup from clusterSnapshot. The necessary columns should be added to the nodeTypeLookup2 and the logic should just be added after the time-series lookup.

another join is not necessary and is wrong as this will create MANY duplicate records since the join could match the same cluster for several times. Example, if we run bronze 3 times before gold this would create 3 duplicate rows for each cluster present in snapshot. This is why time-series lookup is required here.

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
```diff
@@ -279,7 +279,7 @@ trait GoldTransforms extends SparkSessionWrapper {
 
     val derivedSKU = PipelineFunctions.deriveSKU(isAutomated('cluster_name), 'spark_version, 'cluster_type)
     val nodeTypeLookup = clusterSpec.asDF
-      .select('organization_id, 'cluster_id, 'cluster_name, 'custom_tags, 'timestamp, 'driver_node_type_id, 'node_type_id, 'spark_version, 'cluster_type, 'runtime_engine)
+      .select('organization_id, 'cluster_id, 'cluster_name, 'custom_tags, 'timestamp, 'driver_node_type_id, 'node_type_id, 'spark_version, 'cluster_type, 'runtime_engine,'num_workers)
```
GeekSheikh, line 282: This needs to be pulled from clusterSnapshot not clusterSpec. ClusterSpec will also not have the num_workers populated since it's source is also the audit logs. The ticket clearly suggests getting this value from cluster_snapshot not clusterspec; thus this logic should be applied to the `nodeTypeLookup2`, not `nodeTypeLookup` as done here.

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
```diff
@@ -330,12 +330,13 @@ trait GoldTransforms extends SparkSessionWrapper {
           $"clusterPotential.unixTimeMS_state_start"
             .between($"workerNodeDetails.activeFromEpochMillis", $"workerNodeDetails.activeUntilEpochMillis"),
         "left")
-      .drop("activeFrom", "activeUntil", "activeFromEpochMillis", "activeUntilEpochMillis", "activeFromDate", "activeUntilDate", "worker_orgid", "node_type_id_lookup")
+      .withColumn("current_num_workers",when((col("current_num_workers") === "0" && col("num_workers").isNotNull) ,col("num_workers")).otherwise(col("current_num_workers")))
+      .withColumn("target_num_workers",when((col("target_num_workers") === "0" && col("num_workers").isNotNull),col("num_workers")).otherwise(col("target_num_workers")))
```
GeekSheikh, line 334: Are you sure that at this point in the DF current_num_workers will be 0 or will it be null? Also, this is going to adjust ALL records that have current/target_num_workers from 0 to whatever is defined in the snapshot. If null, then we can probably just coalesce in all cases but if it truly is 0 then additional logic is likely necessary to make sure we don't convert true 0s to the snapshot config value.

Overall, I'm concerned that this is not the right place to apply this logic. I believe we should apply the appropriate logic in clusterSpec so that this logic ONLY HAPPENS on FIRST RUN. Doing this every run will likely lead to unexpected overwrites (e.g. single nodes).

GeekSheikh (changes requested):

souravbaner-da:
Currently this change should be put on hold as creating record for the cluster for which we are not getting any cluster event is complex task. May be we would take this up in future release.

For cost discrepancy the issue for long running cluster has been mentioned in Overwatch Doc. Below is the PR for the same
#870

PR #841, opened 2023-03-27
temp_dir for MSW deployment added

### LINKED ISSUE #824: TEMP DIR VIA JAR -- should be configurable
Seems as though the temp directory for the Overwatch run is only editable via a Notebook execution. This should be optionally configurable via Main class as well.

sriram251-code:
closes #824 

PR proof added on GCP workspace.


souravbaner-da (approved):

GeekSheikh (approved):
@sriram251-code -- This looks good but I did notice that the root temp dir is not getting deleted. Not urgent issue but I think we should file a bug to see why the line referenced below is not deleting the parent folder. Could you please file a ticket for 0730 to take another look at this?

https://github.com/databrickslabs/overwatch/blob/5337cadd510f596b03914ea48066db0f5df21641/src/main/scala/com/databricks/labs/overwatch/pipeline/Pipeline.scala#L384

image

PR #843, opened 2023-03-27
Add Helper Function pipReport

souravbaner-da:
Add Helper Function pipReport

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/utils/Tools.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/utils/Tools.scala
@@ -5,20 +5,20 @@ import com.databricks.labs.overwatch.env.Workspace
 import com.databricks.dbutils_v1.DBUtilsHolder.dbutils
 
 import java.io.FileNotFoundException
-import com.databricks.labs.overwatch.pipeline
+import com.databricks.labs.overwatch.pipeline.TransformFunctions._
 import com.databricks.labs.overwatch.pipeline.TransformFunctions.datesStream
 import com.databricks.labs.overwatch.pipeline._
 import com.databricks.labs.overwatch.utils.Helpers.spark.table
 import com.fasterxml.jackson.annotation.JsonInclude.{Include, Value}
 import com.fasterxml.jackson.core.io.JsonStringEncoder
-import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.databind.{JsonNode, ObjectMapper}
 import com.fasterxml.jackson.module.scala.DefaultScalaModule
 import io.delta.tables.DeltaTable
 import org.apache.commons.lang3.StringEscapeUtils
 import org.apache.hadoop.conf._
 import org.apache.hadoop.fs.{FileSystem, Path}
 import org.apache.log4j.{Level, Logger}
-import org.apache.spark.sql.{Dataset, Row}
+import org.apache.spark.sql.{DataFrame, Dataset, Row}
 import org.apache.spark.sql.expressions.Window
 import org.apache.spark.sql.functions._
 import org.apache.spark.util.SerializableConfiguration
@@ -138,10 +138,19 @@ object JsonUtils {
 
 }
 
+trait HelpersMeta{
+  private[overwatch] def getPaginationLogic(jsonObject: JsonNode, requestMap: Map[String, String]): Map[String, String] = {
+    null
+  }
+  private[overwatch] def pipReport(etlDB: String = "" , orgId: String = ""): DataFrame = {
+    null
+  }
+}
+
 /**
  * Helpers object is used throughout like a utility object.
  */
-object Helpers extends SparkSessionWrapper {
+object Helpers extends SparkSessionWrapper with HelpersMeta{
 
   private val logger: Logger = Logger.getLogger(this.getClass)
   private val driverCores = java.lang.Runtime.getRuntime.availableProcessors()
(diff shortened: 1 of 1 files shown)


src/main/scala/com/databricks/labs/overwatch/utils/Tools.scala
```diff
@@ -881,5 +884,28 @@ object Helpers extends SparkSessionWrapper {
...
+
+    val pipReport = if (detailed) {
+      println("Report Extracted from pipeline_report")
+      table(s"${etlDB}.pipeline_report")
+    }else{
+      println("Report Extracted from pipReport")
+      table(s"${etlDB}.pipeline_report")
+    }
+
+    val pipReport1 = if (orgId == "") {
```
sriram251-code, line 897: change variable name

sriram251-code (changes requested):

src/main/scala/com/databricks/labs/overwatch/utils/Tools.scala
```diff
@@ -881,5 +884,38 @@ object Helpers extends SparkSessionWrapper {
...
+        logger.log(Level.ERROR, msg)
+        throw new BadConfigException(msg)
+    }
+
+    val pipReport = if (detailed) {
+      println("Report Extracted from pipeline_report")
+      table(s"${etlDB}.pipeline_report")
+    }else{
+      println("Report Extracted from pipReport")
+      table(s"${etlDB}.pipReport")
```
GeekSheikh, line 903: select only the columns that are necessary and do it from pipeline_report so you don't have to join to get the packageVersion

organization_id, workspace_name, module_id, module_name, status, from_time, until_time, parsedConfig.packageVersion, Pipeline_SnapTS

src/main/scala/com/databricks/labs/overwatch/utils/Tools.scala
```diff
@@ -881,5 +884,38 @@ object Helpers extends SparkSessionWrapper {
...
+    }
 
+    pipReport
+      .orderBy('Pipeline_SnapTS.desc,'moduleID)
+      .withColumn("fromTSt", from_unixtime('fromTS.cast("double") / lit(1000)).cast("timestamp"))
+      .withColumn("untilTSt", from_unixtime('untilTS.cast("double") / lit(1000)).cast("timestamp"))
```
GeekSheikh, line 909: let's change the names to be a little more user friendly for the timestamps, such as "from_time", "until_time".

Let's move the columns in the simple view the front and move the rest of the columns to the end so that it's easier to read.

A select statement with appropriately ordered columns by importance will be very good here. You can either use select statement or df.moveColumnsToFront(...)

src/main/scala/com/databricks/labs/overwatch/utils/Tools.scala
```diff
@@ -881,5 +884,38 @@ object Helpers extends SparkSessionWrapper {
 
   }
 
+  def pipReport(etlDB: String = "overwatch_etl", detailed: Boolean = false):Dataset[Row] = {
```
GeekSheikh, line 887: both signatures should return DataFrame not Dataset[Row] (I know they're basically the same thing but this will get rid of the import and conform to the rest of the codebase.

GeekSheikh (changes requested):

src/main/scala/com/databricks/labs/overwatch/utils/Tools.scala
```diff
@@ -881,5 +884,38 @@ object Helpers extends SparkSessionWrapper {
...
+    pipReport
+      .orderBy('Pipeline_SnapTS.desc,'moduleID)
+      .withColumn("fromTSt", from_unixtime('fromTS.cast("double") / lit(1000)).cast("timestamp"))
+      .withColumn("untilTSt", from_unixtime('untilTS.cast("double") / lit(1000)).cast("timestamp"))
+
+
+  }
+
+
+  def pipReport(etlDB: String = "overwatch_etl", orgId: String = "", detailed: Boolean = false):Dataset[Row] = {
```
GeekSheikh, line 915: you cannot have two functions that implement default parameters. This wouldn't event compile. Always compile and ensure tests pass before submitting PR for review.

src/main/scala/com/databricks/labs/overwatch/utils/Tools.scala
```diff
@@ -881,5 +884,28 @@ object Helpers extends SparkSessionWrapper {
...
   }
 
+  def pipReport(etlDB: String = "overwatch_etl", orgId: String = "", detailed: Boolean = false):Dataset[Row] = {
+
+    val pipReport = if (detailed) {
+      println("Report Extracted from pipeline_report")
+      table(s"${etlDB}.pipeline_report")
+    }else{
+      println("Report Extracted from pipReport")
+      table(s"${etlDB}.pipeline_report")
```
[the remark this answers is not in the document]
GeekSheikh: @sriram251-code -- you cannot overload a function where both functions have defaults set.

GeekSheikh:
Please also add another attribute for functionality --

lastN = -1 (default)

When default -1 -- returns all data
When 1 it uses a window to only grab the latest run for each org/module
When 5 it uses a window to only grab the latest 5 runs for each org/module
When n it uses a window to only grab the latest n runs for each org/module

so basically the number becomes the rank filter

Below is an example of how I use it in the field all the time
```scala
val w = Window.partitionBy('organization_id, 'module_id).orderBy('Pipeline_SnapTS.desc)
display(
  table(s"$TARGETETLDB.pipReport")
    .filter('organization_id === Initializer.getOrgId)
    .withColumn("rnk", rank().over(w))
    .filter('rnk === 1)
    .drop("rnk", "write_metrics")
)
```

GeekSheikh:
There are also conflicts -- be sure to merge 0720_release and/or rebase it then merge it to resolve errors before re-submitting for review.

src/main/scala/com/databricks/labs/overwatch/utils/Tools.scala
```diff
@@ -3,10 +3,12 @@ package com.databricks.labs.overwatch.utils
 import com.amazonaws.services.s3.model.AmazonS3Exception
 import com.databricks.labs.overwatch.env.Workspace
 import com.databricks.dbutils_v1.DBUtilsHolder.dbutils
+
```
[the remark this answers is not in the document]
souravbaner-da: Removed

souravbaner-da:
Raised New PR https://github.com/databrickslabs/overwatch/pull/860

Issue #827, opened 2023-03-23
Handle Exception - Azure ClassDefNotFound - Causes cluster to get stuck forever


GeekSheikh:
In Azure multiworkspace deployments, if customer forgets to attach the Azure Event Hubs JAR the stream will never start but the thread never fails either. There should be an exception handler for this error that fails the entire run and tells the customer to attach the JAR.

The problem presents like this with no streams running and no spark jobs running.
<img width="703" alt="image" src="https://user-images.githubusercontent.com/10840635/227250790-ebdd8ec7-daae-49ce-b459-e3ae9c25a295.png">

Here's what it looks like in the logs.

```
23/03/23 15:01:06 ERROR DatabricksMain$DBUncaughtExceptionHandler: Uncaught exception in thread pool-75-thread-2!
java.lang.NoClassDefFoundError: org/apache/spark/eventhubs/ConnectionStringBuilder$
	at com.databricks.labs.overwatch.pipeline.BronzeTransforms.landAzureAuditLogDF(BronzeTransforms.scala:216)
	at com.databricks.labs.overwatch.pipeline.BronzeTransforms.landAzureAuditLogDF$(BronzeTransforms.scala:207)
	at com.databricks.labs.overwatch.pipeline.Bronze.landAzureAuditLogDF(Bronze.scala:8)
	at com.databricks.labs.overwatch.pipeline.Bronze.landAzureAuditEvents(Bronze.scala:277)
	at com.databricks.labs.overwatch.pipeline.Bronze.$anonfun$executeModules$1(Bronze.scala:291)
	at scala.collection.immutable.List.foreach(List.scala:431)
	at com.databricks.labs.overwatch.pipeline.Bronze.executeModules(Bronze.scala:288)
	at com.databricks.labs.overwatch.pipeline.Bronze.run(Bronze.scala:329)
	at com.databricks.labs.overwatch.MultiWorkspaceDeployment.startBronzeDeployment(MultiWorkspaceDeployment.scala:188)
	at com.databricks.labs.overwatch.MultiWorkspaceDeployment.$anonfun$executePipelines$1(MultiWorkspaceDeployment.scala:414)
	at scala.concurrent.Future$.$anonfun$apply$1(Future.scala:659)
	at scala.util.Success.$anonfun$map$1(Try.scala:255)
	at scala.util.Success.map(Try.scala:213)
	at scala.concurrent.Future.$anonfun$map$1(Future.scala:292)
[...]

PR #848, opened 2023-03-28
handled class not found

sriram251-code:
closing #827 

After the implementation if EH jar is missing then it will break the pipeline as well as it will store the exception in the deployment report.
Screenshot 2023-03-28 at 2 04 47 PM
Screenshot 2023-03-28 at 2 04 27 PM


souravbaner-da (approved):

GeekSheikh (approved):
Works

PR #854, opened 2023-03-30
added safety valve

### LINKED ISSUE #851: SparkStreams - protect against bizarre streams
In some customer accounts -- rare streams can generate literally millions of distinct executionIDs. This is extremely rare but when it happens it significantly impacts cost and performance of operating Overwatch.

Proposed Solution:
Eliminate these streams from gold by adding array_size filter. The ETL will never complete with streams like these and must be eliminated from the streams gold table.

GeekSheikh:
closes #851


souravbaner-da (approved):
Looks Good

PR #859, opened 2023-03-31
getParams implemented

### LINKED ISSUE #857: [FEAT] - MultiWorkspaceDeployment -- add getter for MultiWorkspaceParams
Please add a getter function for MultiWorkspaceDeployment to allow for the following

`MultiWorkspaceDeployment(configPath).getParams()`

getParams func should return MultiWorkspaceParams.

sriram251-code:
closes #857

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/MultiWorkspaceDeployment.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/MultiWorkspaceDeployment.scala
@@ -512,7 +512,7 @@ class MultiWorkspaceDeployment extends SparkSessionWrapper {
    * @param workspaceId
    * @return
    */
-  def getParams(workspaceId: String = "") = {
+  def getParams(workspaceId: String = ""): Array[MultiWorkspaceParams] = {
     val overwatchParams = generateMultiWorkspaceConfig(configLocation, deploymentId, outputPath).map(buildParams)
     val returnParam = if (workspaceId != "") {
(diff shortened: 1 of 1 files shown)


src/main/scala/com/databricks/labs/overwatch/MultiWorkspaceDeployment.scala
```diff
@@ -507,4 +507,19 @@ class MultiWorkspaceDeployment extends SparkSessionWrapper {
     println(msg)
   }
 
+  /**
+   * Returns the Overwatch parameters from config.
+   * @param workspaceId
+   * @return
+   */
+  def getParams(workspaceId: String = "") = {
```
GeekSheikh, line 515: public funcs should always explicitly state return type Array[MultiWorkspaceParams]

GeekSheikh:
Please also add another attribute for functionality --

lastN = -1 (default)

When default -1 -- returns all data
When 1 it uses a window to only grab the latest run for each org/module
When 5 it uses a window to only grab the latest 5 runs for each org/module
When n it uses a window to only grab the latest n runs for each org/module

so basically the number becomes the rank filter

src/main/scala/com/databricks/labs/overwatch/MultiWorkspaceDeployment.scala
sriram251-code: done

GeekSheikh (approved):

PR #860, opened 2023-04-03
Added Logic to create PipReport

### LINKED ISSUE #647: [FEAT] - Add pipReport Helper to display default pipReport
Enable

```scala
Helpers.pipReport(etlDB: string = "overwatch_etl", orgId: String = "*", detailed: Boolean = false)
```

to display the ordered pipReport
```scala
table(s"${etlDB}.pipReport")
  .filter('organization_id === "*")
  .orderBy('Pipeline_SnapTS.desc)
```

if detailed == true use pipeline_report table instead of pipReport view

souravbaner-da:
Added Logic to create PipReport
Close #647

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/utils/Tools.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/utils/Tools.scala
@@ -932,23 +932,24 @@ object Helpers extends SparkSessionWrapper {
 
 
 
-  def pipReport(etlDB: String, noOfRecords:Int, moduleIds: Array[Int], orgId: String*): DataFrame = {
+  def pipReport(
+                 etlDB: String,
+                 noOfRecords:Int = -1,
+                 moduleIds: Array[Int] = Array[Int](),
+                 verbose: Boolean = false,
+                 orgId: Seq[String] = Seq[String]()
+               ): DataFrame = {
     val orgIDFilter = if (orgId.nonEmpty) col("organization_id").isin(orgId: _*) else lit(true)
     val moduleIdFilter = if (moduleIds.nonEmpty) col("moduleId").isin(moduleIds: _*) else lit(true)
-    buildPipReport(etlDB, noOfRecords, orgIDFilter, moduleIdFilter)
+    buildPipReport(etlDB, noOfRecords, orgIDFilter, moduleIdFilter, verbose)
   }
 
-  def pipReport(etlDB: String): DataFrame = {
-    pipReport(etlDB, -1, Array[Int]())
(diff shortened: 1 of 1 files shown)


src/main/scala/com/databricks/labs/overwatch/utils/Tools.scala
```diff
@@ -137,10 +138,28 @@ object JsonUtils {
 
 }
 
+trait HelperTrait{
```
GeekSheikh, line 141: we don't need a trait for this. Adding a trait to the object adds a lot of unnecessary complexity. We can discuss.

GeekSheikh (changes requested):

GeekSheikh:
I have added a suggested / simplified way to do all the things I think you were trying to accomplish. Please review. Also please add docs to all the new public functions

src/main/scala/com/databricks/labs/overwatch/utils/Tools.scala
```diff
@@ -5,9 +5,10 @@ import com.databricks.labs.overwatch.env.Workspace
 import com.databricks.dbutils_v1.DBUtilsHolder.dbutils
 
 import java.io.FileNotFoundException
-import com.databricks.labs.overwatch.pipeline
+import com.databricks.labs.overwatch.pipeline.TransformFunctions._
 import com.databricks.labs.overwatch.pipeline.TransformFunctions.datesStream
 import com.databricks.labs.overwatch.pipeline._
+import com.databricks.labs.overwatch.utils.Helpers.spark.table
```
GeekSheikh, line 11: why are we adding this? What is Helpers.spark.table?

src/main/scala/com/databricks/labs/overwatch/utils/Tools.scala
```diff
@@ -883,5 +884,72 @@ object Helpers extends SparkSessionWrapper {
...
+      spark.catalog.getTable(s"${etlDB}.pipeline_report")
+      logger.log(Level.INFO, s"Overwatch has being deployed with  ${etlDB}")
+    }catch {
+      case e: Exception =>
+        val msg = s"Overwatch has not been deployed with  ${etlDB}"
+        logger.log(Level.ERROR, msg)
+        throw new BadConfigException(msg)
+    }
+
+    val pipReport = table(s"${etlDB}.pipeline_report")
```
GeekSheikh, line 904: use spark.table not `table` so you can get rid of that import.

GeekSheikh (approved):

PR #867, opened 2023-04-06
Implemented AAD for MWS

### LINKED ISSUE #845: Support SP AAD tokens to authenticate against EventHubs for MWS
Now MWS deployment don't support AAD tokens to authenticate against EH. 

Config.csv/delta config should have a column which will be used to support AAD tokens to authenticate agains EH.
Below code is an example of connecting through AAD tokens.
      val basePath = s"$tempPath/$etlDBName"
      val azureLogConfig = AzureAuditLogEventhubConfig(connectionString = evhubConnString, eventHubName = 
     "evhName", auditRawEventsPrefix = basePath,
        azureClientId = Some("<service-principal-client-id>"), // this could be also specified as {{secrets/<scope>/<key>}}
        azureClientSecret = Some("{{secrets/<scope>/<secret>}}"),
        azureTenantId = Some("<azure tenant ID>")
      )

sriram251-code:
closes #845 

PR test is added in the PR folder in Azure workspace.

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/MultiWorkspaceDeployment.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/MultiWorkspaceDeployment.scala
@@ -124,7 +124,7 @@ class MultiWorkspaceDeployment extends SparkSessionWrapper {
           AzureAuditLogEventhubConfig(connectionString = config.eh_conn_string.get, eventHubName = config.eh_name.get
             , auditRawEventsPrefix = ehStatePath,
             azureClientId = Some(config.aad_client_id.get),
-            azureClientSecret = Some(dbutils.secrets.get(scope = "overwatch_global", key = "overwatch-reader-secret")),
+            azureClientSecret = Some(dbutils.secrets.get(config.secret_scope, key = config.aad_client_secret_key.get)),
             azureTenantId = Some(config.aad_tenant_id.get))
         }else{
           val ehConnString = s"{{secrets/${config.secret_scope}/${config.eh_scope_key.get}}}"
--- a/src/main/scala/com/databricks/labs/overwatch/MultiWorkspaceDeployment.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/MultiWorkspaceDeployment.scala
@@ -1,5 +1,6 @@
 package com.databricks.labs.overwatch
 
+import com.databricks.dbutils_v1.DBUtilsHolder.dbutils
 import com.databricks.labs.overwatch.env.Workspace
 import com.databricks.labs.overwatch.pipeline.TransformFunctions._
 import com.databricks.labs.overwatch.pipeline._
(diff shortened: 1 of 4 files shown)


src/main/scala/com/databricks/labs/overwatch/MultiWorkspaceDeployment.scala
```diff
@@ -116,9 +117,21 @@ class MultiWorkspaceDeployment extends SparkSessionWrapper {
...
         val ehStatePath = s"${config.etl_storage_prefix}/${config.workspace_id}/ehState"
-        val azureLogConfig = AzureAuditLogEventhubConfig(connectionString = ehConnString, eventHubName = config.eh_name.get, auditRawEventsPrefix = ehStatePath)
+        val isAAD = config.aad_client_id.nonEmpty && config.aad_tenant_id.nonEmpty && config.aad_client_secret_key.nonEmpty
+        val azureLogConfig = if(isAAD){
+          println("consuming data by AAD")
+          AzureAuditLogEventhubConfig(connectionString = config.eh_conn_string.get, eventHubName = config.eh_name.get
+            , auditRawEventsPrefix = ehStatePath,
+            azureClientId = Some("ba551aff-7ccf-4889-951e-aa491532d844"), // this could be also specified as {{secrets//}}
+            azureClientSecret = Some(dbutils.secrets.get(scope = "overwatch_global", key = "overwatch-reader-secret")),
+            azureTenantId = Some("9f37a392-f0ae-4280-9796-f1864a10effc"))
```
alexott, line 129: why we have it's hardcoded?

src/main/scala/com/databricks/labs/overwatch/pipeline/Schema.scala
```diff
@@ -1062,7 +1062,11 @@ object Schema extends SparkSessionWrapper {
     StructField("enable_unsafe_SSL", BooleanType, nullable = true),
     StructField("thread_pool_size", IntegerType, nullable = true),
     StructField("api_waiting_time", LongType, nullable = true),
-    StructField("mount_mapping_path", StringType, nullable = true)
+    StructField("mount_mapping_path", StringType, nullable = true),
+    StructField("eh_conn_string", StringType, nullable = true),
+    StructField("aad_tenant_id", StringType, nullable = true),
+    StructField("aad_client_id", StringType, nullable = true),
+    StructField("aad_client_secret_key", StringType, nullable = true)
```
alexott, line 1069: we also support AAD authority endpoint (the `aad_authority_endpoint` in the old implementation`) so we can use it not only on "Normal Azure" but also in the GovCloud, China & German Clouds

alexott (changes requested):

src/main/scala/com/databricks/labs/overwatch/pipeline/Schema.scala
sriram251-code: as discussed no code changes needed

src/main/scala/com/databricks/labs/overwatch/MultiWorkspaceDeployment.scala
sriram251-code: my mistake apologies.

src/main/scala/com/databricks/labs/overwatch/pipeline/Schema.scala
```diff
@@ -2,7 +2,7 @@ package com.databricks.labs.overwatch.pipeline
 
 import com.databricks.labs.overwatch.utils.SparkSessionWrapper
 import org.apache.log4j.Logger
-import org.apache.spark.sql.types._
+import org.apache.spark.sql.types.{StructField, _}
```
GeekSheikh, line 5: Please either choose to import all or Structfield, not both

src/main/scala/com/databricks/labs/overwatch/pipeline/Schema.scala
alexott: sorry. Maybe I explained incorrectly - the `aad_authority_endpoint` is necessary if we want to support other Azure clouds: Azure GovCloud, Azure China, Azure Germany

alexott: no. authority endpoint isn't a part of the EH connection string. it's an URL of the AAD REST API that will be used to generate AAD tokens

GeekSheikh:
back-ported to 0714 via #919

PR #872, opened 2023-04-11
AAD authentication for EventHubs implemented for MWS

### LINKED ISSUE #845: Support SP AAD tokens to authenticate against EventHubs for MWS

sriram251-code:
closes #845 
Test notebook "PR_872_AAD_MWS_TEST_720" added in PR test folder.


alexott (changes requested):
as was discussed in another PR, we need to add support for different Azure clouds (make it optional at least).

alexott (approved):
lgtm

src/main/scala/com/databricks/labs/overwatch/validation/DeploymentValidation.scala
```diff
@@ -490,51 +530,22 @@ object DeploymentValidation extends SparkSessionWrapper {
...
-                                scope: String,
-                                optKey: Option[String],
-                                optEHName: Option[String],
-                                outputPath: String
+                               config: MultiWorkspaceConfig
                               ): DeploymentValidationReport = {
-    if (optKey.isEmpty || optEHName.isEmpty) throw new BadConfigException("When cloud is Azure, the eh_name and " +
-      "eh_scope_key are required fields but they were empty in the config")
+
+    val isAAD = config.aad_client_id.nonEmpty && config.aad_tenant_id.nonEmpty && config.aad_client_secret_key.nonEmpty
```
GeekSheikh, line 536: I think you're missing validation of the composite sets.

You need to validate that the config provides one of the two following states and derive isAAD from there -- you should create a stand-alone function for this and provide sufficient unit tests for positive / negative outcomes. If user violates these required conditions -- throw a user friendly error.

isEHAAD:
--- nonEmpty ---
aad_client_id
aad_tenant_id
aad_client_secret_key
eh_name
--- empty ---
eh_scope_key

!isEHAAD
--- nonEmpty ---
eh_scope_key
eh_name
--- empty --- 
aad_client_id
aad_tenant_id
aad_client_secret_key

sriram251-code: done

GeekSheikh (approved):

Issue #882, opened 2023-04-18
API Call Failure in Bronze_SparkEventLogs


gueniai:
When running a Multiworkspace deployment, getting a `None.get` error in module 1006 (Bronze_SparkEventLogs)

Top of stack trace:
`java.util.NoSuchElementException: None.get
	at scala.None$.get(Option.scala:529)
	at scala.None$.get(Option.scala:527)
	at com.databricks.labs.overwatch.pipeline.Initializer$.getOrgId(Initializer.scala:33)
`

PR #883, opened 2023-04-18
Passing the apiURL to getOrgID to make it thread safe

gueniai:
Closes [882](https://github.com/databrickslabs/overwatch/issues/882)

In cases where `Initializer.getOrgId` was executed in the workers, it would result in a `None.get`. This PR overloads the method to create a thread safe alternative. 
Whenever the function was called and the apiURL was already available, I also passed it in, in case we decided to thread those paths in the future.

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/pipeline/Initializer.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/pipeline/Initializer.scala
@@ -27,12 +27,41 @@ object Initializer extends SparkSessionWrapper {
   // Init the SparkSessionWrapper with envVars
   envInit()
 
-  def getOrgId: String = {
+  /**
+   * Returns the local workspace orgID
+   * @return
+   */
+  def getOrgId: String = getOrgId(None)
+
+  /**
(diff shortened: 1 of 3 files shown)


src/main/scala/com/databricks/labs/overwatch/pipeline/Initializer.scala
```diff
@@ -27,13 +27,42 @@ object Initializer extends SparkSessionWrapper {
...
+      val apiURL = getApiURL(apiUrlInput)
+
+      apiURL.split("\\.")(0).split("/").lastOption match {
+        case Some(orgID) => orgID
+        case None => throw new BadConfigException("ORG ID cannot be determined")
+      }
     } else clusterOwnerOrgID
   }
 
+  private def getApiURL(apiUrlInput: Option[String]): String = {
```
GeekSheikh, line 55: In future, try to put helper funcs above usages just a style preference, not wrong.

GeekSheikh, line 55: I like this function -- good improvement over my scribbles

GeekSheikh (approved):

PR #902, opened 2023-04-24
autoOptimize enabled

### LINKED ISSUE #850: [BUG] - AutoOptimize not getting set
When PipelineTable attribute `autoOptimize` is set to true spark configs get set to default writes to create tables with optimizeWrite = true but these settings are not longer taking affect. I can only assume it has to do with the changes made to spark settings for multi-workspace deployment. It's very important that these tables get this property set for performance reasons.

The fact that these are not getting set can be confirmed `desc formatted <table>`. I think we could also add a second step to the "registerTarget" method to alter table and apply these settings when autoOptimize = true to ensure all existing tables get this setting switched.

image

sriram251-code:
closes #850 

PR proof added in the Azure workspace.

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/env/Database.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/env/Database.scala
@@ -27,18 +27,48 @@ class Database(config: Config) extends SparkSessionWrapper {
   }
 
   /**
+   * Checks if the table has optimizeWrite enebled.
+   * @param target
+   * @return
+   */
+  private def checkAutoOptimiseWrite(target: PipelineTable): Boolean = {
+    val deltaProperties = spark.sql(s"""describe detail '${target.tableLocation}'""").select("properties").head().get(0).asInstanceOf[Map[String, Boolean]]
+    val optimiseWrite: Boolean = deltaProperties.get("delta.autoOptimize.optimizeWrite").getOrElse("true").toString.toBoolean
+    optimiseWrite
+  }
+
+
+  /**
    * register an Overwatch target table in the configured Overwatch deployment
    *
    * @param target Pipeline table (i.e. target) as per the Overwatch deployed config
    */
   def registerTarget(target: PipelineTable): Unit = {
     if (!target.exists(catalogValidation = true) && target.exists(pathValidation = true)) {
-      val createStatement = s"create table if not exists ${target.tableFullName} " +
-        s"USING DELTA location '${target.tableLocation}'"
-      val logMessage = s"CREATING TABLE: ${target.tableFullName} at ${target.tableLocation}\n$createStatement\n\n"
-      logger.log(Level.INFO, logMessage)
-      if (config.debugFlag) println(logMessage)
-      spark.sql(createStatement)
+      if (target.autoOptimize) {
+        val isAutoOptimise = checkAutoOptimiseWrite(target)
+        val createStatement = s"create table if not exists ${target.tableFullName} " +
+          s"USING DELTA location '${target.tableLocation}' " +
+          s"TBLPROPERTIES (delta.autoOptimize.optimizeWrite = ${isAutoOptimise},delta.autoOptimize.autoCompact=false)"
+        val logMessage = s"CREATING TABLE: ${target.tableFullName} at ${target.tableLocation}\n$createStatement\n\n"
+        logger.log(Level.INFO, logMessage)
+        if (config.debugFlag) println(logMessage)
+        spark.sql(createStatement) //Creating the table as it was previously created.
+        if (!isAutoOptimise) {
+          val alterStatement = s"""alter table ${target.tableFullName} set TBLPROPERTIES (delta.autoOptimize.optimizeWrite = true)"""
+          logger.log(Level.INFO, alterStatement)
+          if (config.debugFlag) println(alterStatement)
+          spark.sql(alterStatement) //Turning autoOptimize for the table.
+        }
+
+      } else {
+        val createStatement = s"create table if not exists ${target.tableFullName} " +
+          s"USING DELTA location '${target.tableLocation}'"
+        val logMessage = s"CREATING TABLE: ${target.tableFullName} at ${target.tableLocation}\n$createStatement\n\n"
+        logger.log(Level.INFO, logMessage)
+        if (config.debugFlag) println(logMessage)
+        spark.sql(createStatement)
+      }
     }
   }
 
--- a/src/main/scala/com/databricks/labs/overwatch/env/Database.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/env/Database.scala
@@ -27,6 +27,21 @@ class Database(config: Config) extends SparkSessionWrapper {
   }
 
   /**
+   * Checks if the table has optimizeWrite enabled.
+   *
+   * @param target
+   * @return
+   */
+  private def checkAutoOptimizeWrite(target: PipelineTable): Boolean = {
+    import spark.implicits._
+    spark.sql(s"""describe detail '${target.tableLocation}'""")
+      .select("properties").as[Map[String, String]].collect()
+      .headOption.getOrElse(throw new Exception(s"Unable to find properties for ${target.tableLocation}"))
+      .getOrElse("delta.autoOptimize.optimizeWrite", "false")
+      .toBoolean
+  }
+
+  /**
    * It will create the table as it was previously created
    * then check whether it has autoOptimize enabled or not.
    * If AutoOptimize is enabled for the target table but it is not set to true in TBLPROPERTIES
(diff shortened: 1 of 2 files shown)


GeekSheikh:
@sriram251-code I still don't see a proof for this or a requested peer review for this PR. Please get this taken care of and I will review once these are complete.

aman-db (approved):

src/main/scala/com/databricks/labs/overwatch/env/Database.scala
```diff
@@ -26,19 +26,49 @@ class Database(config: Config) extends SparkSessionWrapper {
...
+        val logMessage = s"CREATING TABLE: ${target.tableFullName} at ${target.tableLocation}\n$createStatement\n\n"
+        logger.log(Level.INFO, logMessage)
+        if (config.debugFlag) println(logMessage)
+        spark.sql(createStatement) //Creating the table as it was previously created.
+        if (!isAutoOptimise) {
+          val alterStatement = s"""alter table ${target.tableFullName} set TBLPROPERTIES (delta.autoOptimize.optimizeWrite = true)"""
+          logger.log(Level.INFO, alterStatement)
+          if (config.debugFlag) println(alterStatement)
+          spark.sql(alterStatement) //Turning autoOptimize for the table.
+        }
```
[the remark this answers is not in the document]
sriram251-code: done

src/main/scala/com/databricks/labs/overwatch/env/Database.scala
```diff
@@ -27,18 +27,54 @@ class Database(config: Config) extends SparkSessionWrapper {
...
 
   /**
-   * register an Overwatch target table in the configured Overwatch deployment
+   * Checks if the table has optimizeWrite enabled.
+   * @param target
+   * @return
+   */
+  private def checkAutoOptimizeWrite(target: PipelineTable): Boolean = {
+    import spark.implicits._
+    val optimiseWrite: Boolean = spark.sql(s"""describe detail '${target.tableLocation}'""")
```
GeekSheikh, line 36: sp -- `optimizeWrite`

src/main/scala/com/databricks/labs/overwatch/env/Database.scala
```diff
@@ -27,18 +27,54 @@ class Database(config: Config) extends SparkSessionWrapper {
...
+      .select("properties").as[Map[String, String]].collect()
+      .headOption.getOrElse(throw new Exception("string"))
```
GeekSheikh, line 38: put a valid exception string here not `string`

src/main/scala/com/databricks/labs/overwatch/env/Database.scala
```diff
@@ -27,18 +27,54 @@ class Database(config: Config) extends SparkSessionWrapper {
...
+      .getOrElse("delta.autoOptimize.optimizeWrite", "false")
+      .toBoolean
+    optimiseWrite
+  }
+
+
+  /**
+   * register an Overwatch target table in the configured Overwatch deployment.
+   * After release > 7.2.1 : When the tables are dropped ,it will create the table as it was previously created then check whether it has autoOptimize enabled or not.
```
GeekSheikh, line 47: As of 0.7.2.1 -- not after 0721

src/main/scala/com/databricks/labs/overwatch/env/Database.scala
```diff
@@ -27,18 +27,54 @@ class Database(config: Config) extends SparkSessionWrapper {
...
+   * If AutoOptimize is enables then alter the table and set the property of autoOptimizeWrite to true.
```
GeekSheikh, line 48: could you be little more explicit on the rules here. Please also note that creating a table in a way that does not match the underlying files will result in an error. If you write out the logic in English, you'll realize that the code logic does not satisfy the intent. See below.

src/main/scala/com/databricks/labs/overwatch/env/Database.scala
```diff
@@ -27,18 +27,54 @@ class Database(config: Config) extends SparkSessionWrapper {
...
+        if (config.debugFlag) println(logMessage)
+        spark.sql(createStatement) //Creating the table as it was previously created.
+        if (!targetPropertiesAccurate) {
+          val alterStatement = s"""alter table ${target.tableFullName} set TBLPROPERTIES (delta.autoOptimize.optimizeWrite = true)"""
+          logger.log(Level.INFO, alterStatement)
+          if (config.debugFlag) println(alterStatement)
+          spark.sql(alterStatement) //Turning autoOptimize for the table.
+        }
+
+      } else {
```
GeekSheikh, line 70: Please add a comment to this else

```scala
} else { // target is not auto optimize
```

src/main/scala/com/databricks/labs/overwatch/env/Database.scala
```diff
@@ -27,18 +27,54 @@ class Database(config: Config) extends SparkSessionWrapper {
...
-        s"USING DELTA location '${target.tableLocation}'"
-      val logMessage = s"CREATING TABLE: ${target.tableFullName} at ${target.tableLocation}\n$createStatement\n\n"
-      logger.log(Level.INFO, logMessage)
-      if (config.debugFlag) println(logMessage)
-      spark.sql(createStatement)
+      if (target.autoOptimize) {//Checks if the target table should be autoOptimize or not.
+        val targetPropertiesAccurate = checkAutoOptimizeWrite(target)
+        val createStatement = s"create table if not exists ${target.tableFullName} " +
+          s"USING DELTA location '${target.tableLocation}' " +
+          s"TBLPROPERTIES (delta.autoOptimize.optimizeWrite = ${targetPropertiesAccurate},delta.autoOptimize.autoCompact=false)"
```
GeekSheikh, line 58: you should force this boolean to a string

`${targetPropertiesAccurate.toString}`

src/main/scala/com/databricks/labs/overwatch/env/Database.scala
```diff
@@ -27,18 +27,54 @@ class Database(config: Config) extends SparkSessionWrapper {
...
+        val logMessage = s"CREATING TABLE: ${target.tableFullName} at ${target.tableLocation}\n$createStatement\n\n"
+        logger.log(Level.INFO, logMessage)
+        if (config.debugFlag) println(logMessage)
+        spark.sql(createStatement) //Creating the table as it was previously created.
+        if (!targetPropertiesAccurate) {
+          val alterStatement = s"""alter table ${target.tableFullName} set TBLPROPERTIES (delta.autoOptimize.optimizeWrite = true)"""
```
GeekSheikh, line 64: this should not ALWAYS be set to true if `!targetPropertiesAccurate`. What if optimizeWrite in the table is true but it's definition defines that it should be false?

I realize that you're in an if block to handle only target.autoOptimize but nonetheless, the code is confusing and you don't handle if the opposite is true (config == false and actual == true)

src/main/scala/com/databricks/labs/overwatch/env/Database.scala
```diff
@@ -27,18 +27,54 @@ class Database(config: Config) extends SparkSessionWrapper {
...
+   */
+  private def checkAutoOptimizeWrite(target: PipelineTable): Boolean = {
+    import spark.implicits._
+    val optimiseWrite: Boolean = spark.sql(s"""describe detail '${target.tableLocation}'""")
+      .select("properties").as[Map[String, String]].collect()
+      .headOption.getOrElse(throw new Exception("string"))
+      .getOrElse("delta.autoOptimize.optimizeWrite", "false")
+      .toBoolean
+    optimiseWrite
+  }
```
GeekSheikh, line 42: seems like this function should not return whether or not the property is true or false but rather whether or not it matches the target config, right?

shouldn't it return
`optimizeWrite == target.autoOptimize`?

I think you're trying to get too fancy with what this function does. Perhaps two functions would be easier to follow. 

1. Is the target actually optimizeWrite
2. Does the target actual setting match it's config

src/main/scala/com/databricks/labs/overwatch/env/Database.scala
GeekSheikh, line 64: Since you're inside an if block of `!target.exists(catalogValidation = true)` then this will never work because the table does not exist once you get to this block of code since alter table tableName will result in Table Does not Exist.

GeekSheikh (changes requested):
I think you need to refactor the entire registerTarget function to handle all the cases. Perhaps you should first write down all the cases that are to be handled. Refactor this into a few functions if necessary to handle all the scenarios. Develop sufficient tests to ensure that all scenarios are handled.

Please update the function definition and improve your proof to account for the scenarios you intend to handle.

GeekSheikh (approved):

PR #917, opened 2023-05-02
[FEAT] - Verbose audit logging 721

### LINKED ISSUE #622: [FEAT] - Verbose Audit Logging - Gold View or Target
With the release of [verbose audit logging feature](https://docs.databricks.com/administration-guide/account-settings/audit-logs.html#configure-verbose-audit-logs), OW captures all of this data in bronze. It's quite simple to expose this via a view to provide this information to customers. Here's the [internal logic](https://adb-2222170229861029.9.azuredatabricks.net/?o=2222170229861029#notebook/3319176468439680/command/3319176468439681) on how this can be exposed as a gold level view.

souravbaner-da:
Close #622


brij-raghuwanshi-db (approved):
Went through the Proof. Looks Good.

src/main/scala/com/databricks/labs/overwatch/pipeline/Gold.scala
```diff
@@ -359,10 +373,41 @@ class Gold(_workspace: Workspace, _database: Database, _config: Config)
...
+          case OverwatchScope.notebooks =>
+            config.overwatchScope.foreach {
+              case OverwatchScope.clusterEvents => {
+                notebookCommandsModule.execute(appendNotebookCommandsProcess)
+                GoldTargets.notebookCommandsTargetView.publish(verboseAuditTargetViewColumnMapping)
+              }
+              case _ =>
+            }
+          case _ =>
+        }
```
GeekSheikh, line 387: Please create complex boolean to determine if verboseAuditlogging should be processed and then reference this in a single top level command -- I don't want to nest these case statements like this.

Also make it accessible from both locations where nested cases are used.

src/main/scala/com/databricks/labs/overwatch/pipeline/Gold.scala
```diff
@@ -286,6 +286,20 @@ class Gold(_workspace: Workspace, _database: Database, _config: Config)
       )
   }
 
+  lazy private[overwatch] val notebookCommandsModule = Module(3018, "Gold_NotebookCommands", this, Array(1004,3004,3005))
+  lazy private val appendNotebookCommandsProcess: () => ETLDefinition = {
+    () =>
+      ETLDefinition(
+        BronzeTargets.auditLogsTarget.asIncrementalDF(notebookCommandsModule, BronzeTargets.auditLogsTarget.incrementalColumns),
```
GeekSheikh, line 293: need to add 1 day lag for commands that arrive late into the audit bronze source.

src/main/scala/com/databricks/labs/overwatch/pipeline/PipelineTargets.scala
```diff
@@ -498,6 +498,21 @@ abstract class PipelineTargets(config: Config) {
       config
     )
 
+    lazy private[overwatch] val notebookCommandsTarget: PipelineTable = PipelineTable(
+      name = "notebookCommands_gold",
+      _keys = Array("organization_id", "unixTimeMS"),
+      config,
+      _mode = WriteMode.append,
```
GeekSheikh, line 505: Should be merge - insert only

src/main/scala/com/databricks/labs/overwatch/pipeline/PipelineTargets.scala
```diff
@@ -498,6 +498,21 @@ abstract class PipelineTargets(config: Config) {
       config
     )
 
+    lazy private[overwatch] val notebookCommandsTarget: PipelineTable = PipelineTable(
+      name = "notebookCommands_gold",
+      _keys = Array("organization_id", "unixTimeMS"),
```
GeekSheikh, line 503: notebook_id must be part of the key.

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
```diff
@@ -493,6 +493,67 @@ trait GoldTransforms extends SparkSessionWrapper {
...
+    val colNames = Array(
+      "organization_id", "cluster_id","notebook_id", "workspace_name", "date", "timestamp",
+       "notebook_path", "notebook_name", "command_id", "command_text", "execution_time_s", "source_ip_address",
+      "user_identity", "estimated_dbu_cost", "status", "cluster_name", "custom_tags",
+      "node_type_id", "node_count", "response", "user_agent", "unixTimeMS"
+    )
+
+    val notebookCodeAndMetaDF = auditIncrementalDF
+      .filter(col("serviceName") === "notebook" && col("actionName") === "runCommand")
+      .verifyMinimumSchema(Schema.auditMasterSchema)
```
GeekSheikh, line 529: This module_id should be added to Schema.scala and this verification should be done on module start through the standard process -- don't verifyMinSchema here.

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
```diff
@@ -493,6 +493,67 @@ trait GoldTransforms extends SparkSessionWrapper {
...
+      .selectExpr("*", "requestParams.*").drop("requestParams")
+      .withColumnRenamed("timestamp", "unixTimeMS")
```
GeekSheikh, line 531: Since notebook_id must be a key -- be sure to filter where it's not null.

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
```diff
@@ -493,6 +493,67 @@ trait GoldTransforms extends SparkSessionWrapper {
...
+                                   notebook: PipelineTable,
+                                   clsfIncrementalDF : DataFrame,
+                                 )(auditIncrementalDF: DataFrame): DataFrame = {
+
+
+    val notebookLookupTSDF = notebook.asDF
+      .select("organization_id", "notebook_id", "notebook_path", "notebook_name", "unixTimeMS","date")
+      .withColumnRenamed("notebook_id", "notebookId")
+      .distinct
+      .toTSDF("unixTimeMS", "organization_id","date","notebookId")
```
GeekSheikh, line 507: organization_id should come first as it's the most inclusive key and partition filter.

GeekSheikh: add filter to remove null notebook_ids if there are any.

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
```diff
@@ -493,6 +493,67 @@ trait GoldTransforms extends SparkSessionWrapper {
...
+    )
+
+    val notebookCodeAndMetaDF = auditIncrementalDF
+      .filter(col("serviceName") === "notebook" && col("actionName") === "runCommand")
+      .verifyMinimumSchema(Schema.auditMasterSchema)
+      .selectExpr("*", "requestParams.*").drop("requestParams")
+      .withColumnRenamed("timestamp", "unixTimeMS")
+      .withColumn("timestamp", from_unixtime(col("unixTimeMS") / 1000.0).cast("timestamp"))
+      .withColumn("executionTime",col("executionTime").cast("double"))
+      .toTSDF("unixTimeMS", "organization_id", "date","notebookId")
```
GeekSheikh, line 534: switch organization_id to be first

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
```diff
@@ -493,6 +493,67 @@ trait GoldTransforms extends SparkSessionWrapper {
...
+      .lookupWhen(notebookLookupTSDF)
+      .df
+      .toTSDF("unixTimeMS", "organization_id", "clusterId")
```
GeekSheikh, line 537: switch organization_id to be first

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
```diff
@@ -493,6 +493,67 @@ trait GoldTransforms extends SparkSessionWrapper {
...
+
+    val notebookCodeAndMetaDF = auditIncrementalDF
+      .filter(col("serviceName") === "notebook" && col("actionName") === "runCommand")
+      .verifyMinimumSchema(Schema.auditMasterSchema)
+      .selectExpr("*", "requestParams.*").drop("requestParams")
+      .withColumnRenamed("timestamp", "unixTimeMS")
+      .withColumn("timestamp", from_unixtime(col("unixTimeMS") / 1000.0).cast("timestamp"))
+      .withColumn("executionTime",col("executionTime").cast("double"))
+      .toTSDF("unixTimeMS", "organization_id", "date","notebookId")
+      .lookupWhen(notebookLookupTSDF)
```
GeekSheikh, line 535: you might get more notebook_paths / meta if you allow a lookAhead here -- suggest setting lookAhead to like 100 rows or something -- lookback can probably remain max since we shouldn't ever have an insane number of records in a single run.

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
```diff
@@ -493,6 +493,67 @@ trait GoldTransforms extends SparkSessionWrapper {
...
+      .df
+      .toTSDF("unixTimeMS", "organization_id", "clusterId")
+      .lookupWhen(clsfLookupTSDF)
```
GeekSheikh, line 538: some notebooks will run on clusters with massive number of events -- so you need to add in a tsPartitionVal of like 8 or something to handle skewed windows.

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
```diff
@@ -493,6 +493,67 @@ trait GoldTransforms extends SparkSessionWrapper {
...
+      .withColumnRenamed("current_num_workers", "node_count")
+      .withColumn("totalCostPMS", col("total_cost") / col("uptime_in_state_H"))
+      .toTSDF("unixTimeMS", "organization_id", "clusterId")
+
+    val colNames = Array(
+      "organization_id", "cluster_id","notebook_id", "workspace_name", "date", "timestamp",
+       "notebook_path", "notebook_name", "command_id", "command_text", "execution_time_s", "source_ip_address",
+      "user_identity", "estimated_dbu_cost", "status", "cluster_name", "custom_tags",
+      "node_type_id", "node_count", "response", "user_agent", "unixTimeMS"
+    )
```
GeekSheikh, line 525: convert this to an Array[Column] and add all your aliasing here so you don't have to have so many .withColumnRenamed rows. .withColumnRenamed and .withColumn both add a nested spark plan and causes much larger plan deser time -- this is the purpose of this pattern throughout the code.

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
```diff
@@ -493,6 +493,67 @@ trait GoldTransforms extends SparkSessionWrapper {
...
+      .withColumnRenamed("executionTime","execution_time_s")
+      .withColumnRenamed("sourceIpAddress","source_ip_address")
+      .withColumnRenamed("userIdentity","user_identity")
+      .drop("cluster_id")
+      .withColumnRenamed("clusterId","cluster_id")
+      .withColumnRenamed("userAgent","user_agent")
+
+    println("Verbose Audit Log Created")
+
+    notebookCodeAndMetaDF.select(colNames map col: _*)
```
GeekSheikh, line 554: this .select should be selecting all the column names such that you don't need all the rename calls. move this up and remove all the renames.

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
```diff
@@ -493,6 +493,67 @@ trait GoldTransforms extends SparkSessionWrapper {
...
+      .withColumn("estimated_dbu_cost", col("executionTime") * col("totalCostPMS"))
+      .withColumnRenamed("user_identity","userIdentity")
+      .withColumnRenamed("notebookId","notebook_id")
+      .withColumnRenamed("commandId","command_id")
+      .withColumnRenamed("commandText","command_text")
+      .withColumnRenamed("executionTime","execution_time_s")
+      .withColumnRenamed("sourceIpAddress","source_ip_address")
+      .withColumnRenamed("userIdentity","user_identity")
+      .drop("cluster_id")
+      .withColumnRenamed("clusterId","cluster_id")
```
GeekSheikh, line 549: be careful here.

GeekSheikh (changes requested):
This looks MUCH better! Still some changes (much smaller than before) please let me know if you have any questions.

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
souravbaner-da: Added

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
souravbaner-da: Added

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
souravbaner-da: Added

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
souravbaner-da: Done

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
souravbaner-da: As per the toTSDF signature the first argument should be some ts column. So should it not be 'unixTimeMs" the first argument here. I can see same has been applied to other modules too . @GeekSheikh  correct me if I am wrong here

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
souravbaner-da: As per the toTSDF signature the first argument should be some ts column. So should it not be 'unixTimeMs" the first argument here. I can see same has been applied to other modules too . @GeekSheikh  correct me if I am wrong here

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
souravbaner-da: Added the filter

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
souravbaner-da: Agreed. Added

src/main/scala/com/databricks/labs/overwatch/pipeline/PipelineTargets.scala
souravbaner-da: Agreed . Added

src/main/scala/com/databricks/labs/overwatch/pipeline/PipelineTargets.scala
souravbaner-da: Added

src/main/scala/com/databricks/labs/overwatch/pipeline/Gold.scala
souravbaner-da: Agree. Added

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
souravbaner-da: Agreed. This actually already handled in Execute function

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
```diff
@@ -493,6 +493,91 @@ trait GoldTransforms extends SparkSessionWrapper {
...
+        maxLookAhead = 100
+      )
+      .df
+      .toTSDF("unixTimeMS", "organization_id", "clusterId")
+      .lookupWhen(
+        clsfLookupTSDF,
+        tsPartitionVal = 8
+      )
+      .df
+      .withColumn("estimated_dbu_cost", col("executionTime") * col("totalCostPMS"))
```
GeekSheikh, line 565: looks like there's an issue here

something that runs on 8 nodes for 1275 seconds CANNOT cost $30K. ;) Please add some logic to your proof to do a sanity check on estimated costs.

image

GeekSheikh (changes requested):
Couple of issues.

In your proof, please make sure to add the following:
* sanity check logic for costs
* sanity checks for runtimes and start / end timestamps
* a second pipeline run to prove that first and subsequent runs are populated as expected
* prove that the view is present and operational
* desc formatted on the table to show structure
* show create view on the view to show DDL for create

brij-raghuwanshi-db (approved):
Verified with the data and the supporting notebook.
Looks good.
IT now has command-level logging and DBU spends information.

src/main/scala/com/databricks/labs/overwatch/pipeline/Gold.scala
```diff
@@ -89,6 +89,11 @@ class Gold(_workspace: Workspace, _database: Database, _config: Config)
   private val clsfSparkOverrides = Map(
     "spark.sql.adaptive.skewJoin.skewedPartitionThresholdInBytes" -> "67108864" // lower to 64MB due to high skew potential
   )
+
+  private val  verboseAuditLoggingDerivedScope: Boolean = config.overwatchScope.contains(OverwatchScope.audit) && config.overwatchScope.contains(OverwatchScope.notebooks) && config.overwatchScope.contains(OverwatchScope.clusterEvents)
```
GeekSheikh, line 93: This line is too long, please add appropriate line breaks.

src/main/scala/com/databricks/labs/overwatch/pipeline/PipelineTargets.scala
```diff
@@ -498,6 +498,22 @@ abstract class PipelineTargets(config: Config) {
       config
     )
 
+    lazy private[overwatch] val notebookCommandsTarget: PipelineTable = PipelineTable(
```
GeekSheikh, line 501: you added a target and a module. You need to add them to the Gold target / module inventory.

Gold.scala
`getAllTargets`
`getAllModules`

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
```diff
@@ -493,6 +493,82 @@ trait GoldTransforms extends SparkSessionWrapper {
...
+      'clusterId.alias("cluster_id"),
+      'notebookId.alias("notebook_id"),
+      'workspace_name,
+      'date,
+      'timestamp,
+      'notebook_path,
+      'notebook_name,
+      'commandId.alias("command_id"),
+      'commandText.alias("command_text"),
+      'executionTime.alias("execution_time_s"),
```
GeekSheikh, line 532: did we confirm that these are in seconds?

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
```diff
@@ -493,6 +493,82 @@ trait GoldTransforms extends SparkSessionWrapper {
...
+
+    val clsfLookupTSDF = clsfIncrementalDF
+      .select(
+        "organization_id", "state_start_date", "unixTimeMS_state_start", "cluster_id", "cluster_name",
+        "custom_tags", "node_type_id", "current_num_workers", "uptime_in_state_H", "total_cost")
+      .distinct
+      .withColumnRenamed("unixTimeMS_state_start", "unixTimeMS")
+      .withColumnRenamed("cluster_id", "clusterId")
+      .withColumnRenamed("current_num_workers", "node_count")
+      .withColumn("totalCostPMS", col("total_cost") / col("uptime_in_state_H")/lit(3600))
```
GeekSheikh, line 518: total_cost shoudl be total_dbu_cost since you're creating field `estimated_dbu_cost`. total_cost includes compute costs which we don't want for this.

GeekSheikh, line 518: totalCostPMS ?? What is this? I thought this was totalCost per millisecond, is that right? If so why are you dividing hours by 3600? Furthermore, if it's millis it should be Hour * 3600 * 1000, right?

If the `executionTime` is in seconds, then you want this in seconds at which point hours * 3600 would be to the seconds but then you should change the name to be total_dbu_cost_ps or something similar to notate that it's in seconds.

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
```diff
@@ -493,6 +493,82 @@ trait GoldTransforms extends SparkSessionWrapper {
...
+      'date,
+      'timestamp,
+      'notebook_path,
+      'notebook_name,
+      'commandId.alias("command_id"),
+      'commandText.alias("command_text"),
+      'executionTime.alias("execution_time_s"),
+      'sourceIpAddress.alias("source_ip_address"),
+      'userIdentity.alias("user_identity"),
+      'estimated_dbu_cost,
```
GeekSheikh, line 535: In your proof you should add a sanity check for this. Find a command that ran for some reasonable amount of time (longer is better to get bigger numbers), then do a manual calculation of how much you think it should cost based on the time cost of that clusters state while it ran and make sure what you're getting for this field is fairly close to it. I don't think this calc is right yet.

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
```diff
@@ -493,6 +493,82 @@ trait GoldTransforms extends SparkSessionWrapper {
...
+      .lookupWhen(
+        notebookLookupTSDF,
+        maxLookAhead = 100
+      )
+      .df
+      .toTSDF("unixTimeMS", "organization_id", "clusterId")
+      .lookupWhen(
+        clsfLookupTSDF,
+        tsPartitionVal = 8
+      )
```
GeekSheikh, line 563: this is not the right way to derive this. 

Imagine you have a command that runs for 2 hours and the cluster changes states 4 times during that time...the cost per second will be different at different states throughout the runtime of this cmd execution. You need to complete something like what is below to get the cost per run state. I will DM you a link to a video that will help you understand how to build and work with runstates.

See WorkflowTransforms.scala funcs `jrcpDeriveRunsByClusterState` & `jrcpDeriveCumulativeRuntimeByRunState`.  You need to calculate the costs at the various states of the cluster across the cmdExecution, aggregate the per milli cost across each state to calculate the cost / coreMillis of a cmd execution.

Currently, your logic simply grabs the state at the start of the command but doesn't account for state changes during cmd execution.

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
```diff
@@ -493,6 +493,82 @@ trait GoldTransforms extends SparkSessionWrapper {
...
+      'organization_id,
+      'clusterId.alias("cluster_id"),
+      'notebookId.alias("notebook_id"),
+      'workspace_name,
+      'date,
+      'timestamp,
+      'notebook_path,
+      'notebook_name,
+      'commandId.alias("command_id"),
+      'commandText.alias("command_text"),
```
GeekSheikh, line 531: currently, if a customer doesn't have verboseAuditLogs enabled, these fields will be missing and the fact that you're selecting these columns will break the pipeline. ALL NEW FIELDS must be added to audit_log_bronze appropriately in Schema.scala. Please double check that all fields are there and typed correctly (careful not to duplicate).

GeekSheikh (dismissed):
I'm super excited to see this come to gold, what a powerful table! That said, the numbers aren't quite right yet, see comments for more details. We've still got a ways to go to get this to be accurate and ready to merge.

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
```diff
@@ -493,6 +493,149 @@ trait GoldTransforms extends SparkSessionWrapper {
...
+      .filter('notebookId.isNotNull)
+      .distinct
+      .toTSDF("unixTimeMS", "organization_id","date","notebookId")
+
+    val clsfDF = clsfIncrementalDF
+      .select(
+        "organization_id", "state_start_date", "unixTimeMS_state_start", "cluster_id",
+        "current_num_workers", "uptime_in_state_H", "total_DBU_cost","unixTimeMS_state_end")
+      .distinct
+      //      .withColumnRenamed("unixTimeMS_state_start", "unixTimeMS")
```
gueniai, line 516: remove unused code

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
```diff
@@ -493,6 +493,149 @@ trait GoldTransforms extends SparkSessionWrapper {
...
+    // cluster_state started after command start time and ended after command end time
+    val state_after = 'unixTimeMS_state_start > 'unixTimeMSStart && 'unixTimeMS_state_end > 'unixTimeMSEnd
+
+    // Cluster State Started before cmd start time and ended after command run time
+    val state_before_after = 'unixTimeMS_state_start < 'unixTimeMSStart && 'unixTimeMS_state_end > 'unixTimeMSEnd
+
+    val windowSpec  = Window.partitionBy("commandId","clusterId","unixTimeMSStart","unixTimeMSEnd").orderBy("executionTime")
+
+    val joinedDF_before_before = joinedDF.filter(state_before_before)
+      .withColumn("avg_total_dbu_cost_ps", avg(col("total_dbu_cost_ps")).over(windowSpec))
```
gueniai, line 571: an average of averages will almost never yield the right numbers, and it will be incorrect in this case. Please surface the raw values from CLSF and do sums along the way, and only do an avg at the very end.

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
```diff
@@ -493,6 +493,149 @@ trait GoldTransforms extends SparkSessionWrapper {
...
+      .toTSDF("unixTimeMS", "organization_id", "date","notebookId")
+            .lookupWhen(
+              notebookLookupTSDF,
+              maxLookAhead = 100
+            )
+      .df
+      .withColumn("estimated_dbu_cost", col("executionTime") * col("total_dbu_cost_ps"))
+      .drop("cluster_id")
+
+    println("Verbose Audit Log Created")
```
gueniai, line 611: I don't think it's necessary to print here

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
```diff
@@ -493,6 +493,149 @@ trait GoldTransforms extends SparkSessionWrapper {
...
+    val notebookClsfDF = unionDF.union(df.select(unionDF.schema.fieldNames.map(col): _*))
+    val notebookCodeAndMetaDF = notebookClsfDF
+      .toTSDF("unixTimeMS", "organization_id", "date","notebookId")
+            .lookupWhen(
+              notebookLookupTSDF,
+              maxLookAhead = 100
+            )
+      .df
+      .withColumn("estimated_dbu_cost", col("executionTime") * col("total_dbu_cost_ps"))
+      .drop("cluster_id")
```
gueniai, line 609: why don't we include the cluster_id in the final dataframe?

src/main/scala/com/databricks/labs/overwatch/utils/Tools.scala
```diff
@@ -1029,4 +1029,24 @@ object Helpers extends SparkSessionWrapper {
     when(url.endsWith("/"), url.substr(lit(0), length(url) - 1)).otherwise(url)
   }
 
+  /**
```
gueniai, line 1032: I don't think this is part of your changes, right?

gueniai (changes requested):
Left you some comments, let me know if you have any questions. Pointed out a few comments from Daniel's latest review that are still pending.
One point for you to do one full pass is that we shouldn't have any references to "verbose logging" in the code. This is just the means how we're building this table, but this PR is all about building the Notebook Commands table.

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
```diff
@@ -493,6 +493,143 @@ trait GoldTransforms extends SparkSessionWrapper {
...
+    val notebookClsfDF = unionDF_final.union(df.select(unionDF_final.schema.fieldNames.map(col): _*))
+
+    val notebookCodeAndMetaDF = notebookClsfDF.withColumnRenamed("unixTimeMS_state_start","unixTimeMS")
+      .toTSDF("unixTimeMS", "organization_id", "date","notebookId")
+            .lookupWhen(
+              notebookLookupTSDF,
+              maxLookAhead = 100
+            )
+      .df
+      .withColumn("total_dbu_cost_ps", col("total_DBU_cost") / col("uptime_in_state_H")/lit(3600))
```
gueniai, line 626: Semantically, this is not really total_dbu_cost_ps, this is more like average dbu cost per second, or just dbu_cost_ps, right?

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
gueniai: still confused by this one

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
gueniai: reminder to do this sanity check ^

gueniai (changes requested):

src/main/scala/com/databricks/labs/overwatch/pipeline/Gold.scala
```diff
@@ -74,6 +79,11 @@ class Gold(_workspace: Workspace, _database: Database, _config: Config)
       }
       case _ => Array[Module]()
     }
+    if(notebookCommandsDerivedScope) {
```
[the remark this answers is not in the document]
souravbaner-da: Done @gueniai

src/main/scala/com/databricks/labs/overwatch/pipeline/PipelineTargets.scala
```diff
@@ -518,6 +518,22 @@ abstract class PipelineTargets(config: Config) {
...
+      name = "notebookCommands_gold",
+      _keys = Array("notebook_id", "unixTimeMS"),
```
gueniai, line 523: shouldn't the command_id also be a key of this table?

src/main/scala/com/databricks/labs/overwatch/pipeline/Gold.scala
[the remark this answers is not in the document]
gueniai: I don't see any changes to the code @souravbaner-da

gueniai (changes requested):

src/main/scala/com/databricks/labs/overwatch/pipeline/PipelineTargets.scala
souravbaner-da: umm but command_id would be having unique value for each row and if we make it as key wouldn't it make the indexing slower?

gueniai: I was thinking more of what if we have multiple very small commands that come in at the exact same unixTimeMS, we would end up with only the first one we see, no? But walking through the scenario I think it's not very likely, since this is at the millisecond level.

gueniai (approved):
Done!

PR #934, opened 2023-05-16
[Bug] Permanently deleted clusters should stop accumulating cost

### LINKED ISSUE #846: [BUG] - Permanently Deleted Clusters should stop accumulating cost
When a cluster is permanently deleted it and all its data (including event history) is also deleted. As such when the subsequent OW run executes it no longer sees cluster state changes and thinks it's still in the same state as it was during the last run and thus appends charges.

Proposal -- when the permanentDelete event is captured in the auditLog close the outstanding state at the time of the delete event. This still won't be perfectly accurate but it will be much better than accumulating costs for 30 days after the delete.

This logic should be added to cluster_state_detail_silver and/or clsf.

souravbaner-da:
Close #846

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
@@ -1125,9 +1125,6 @@ trait SilverTransforms extends SparkSessionWrapper {
           lead('timestamp, 1).over(stateUnboundW) - lit(1), // subtract 1 millis
           lit(untilTime.asUnixTimeMilli)
         )))
-
-
-
 //      .withColumn("unixTimeMS_state_start", 'timestamp)
 //      .withColumn("unixTimeMS_state_end", coalesce( // if state end open, use pipelineSnapTime, will be merged when state end is received
 //        lead('timestamp, 1).over(stateUnboundW) - lit(1), // subtract 1 millis
--- a/src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
@@ -1098,6 +1098,7 @@ trait SilverTransforms extends SparkSessionWrapper {
       .withColumn("reset_partition", sum('counter_reset).over(stateUnboundW))
       .withColumn("target_num_workers", last('target_num_workers, true).over(stateUnboundW))
       .withColumn("current_num_workers", last('current_num_workers, true).over(stateUnboundW))
+      .withColumn("unixTimeMS_state_start", 'timestamp)
       //Change for Issue-846
 
     val clusterBaseDF = clusterBase(auditLogDF)
--- a/src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
@@ -986,7 +986,8 @@ trait SilverTransforms extends SparkSessionWrapper {
   }
 
   def buildClusterStateDetail(
-                               untilTime: TimeTypes
+                               untilTime: TimeTypes,
+                               auditLogDF: DataFrame
                              )(clusterEventsDF: DataFrame): DataFrame = {
     val stateUnboundW = Window.partitionBy('organization_id, 'cluster_id).orderBy('timestamp)
     val stateFromCurrentW = Window.partitionBy('organization_id, 'cluster_id).rowsBetween(1L, 1000L).orderBy('timestamp)
@@ -1086,7 +1087,7 @@ trait SilverTransforms extends SparkSessionWrapper {
         'organization_id, 'cluster_id, 'isRunning,
         'timestamp, 'state, 'current_num_workers, 'target_num_workers
       )
-
+    val clusterEventsBaselineFinal =
     clusterEventsBaseline
       .withColumn("counter_reset",
         when(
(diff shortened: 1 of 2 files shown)


src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
```diff
@@ -1098,23 +1099,50 @@ trait SilverTransforms extends SparkSessionWrapper {
...
+
+    val removedClusterID = clustersRemoved.select("cluster_id").distinct().collect().map(f=>f(0))
+    val w = Window.partitionBy('organization_id, 'cluster_id).orderBy('unixTimeMS_state_start)
+    val timeStampForRemovedCluster = clusterEventsBaselineFinal
+      .drop("driverSpecs", "workerSpecs")
+      .filter(col("cluster_id").isin(removedClusterID: _*))
+      .withColumn("positiveTestCase", when(last('state).over(w) === "RUNNING", lit(true)).otherwise(lit(false)))
+      .withColumn("positiveTestCase2", lead('positiveTestCase,1).over(w))
+      .filter('positiveTestCase2.isNull)
+      .filter('state === "RUNNING").select("timestamp").collect().map(t=>t(0))
```
GeekSheikh, line 1117: why do all of this? There's a field later that is just `isRunning` if you move this logic to the end of this function, can you not just use the already-derived `isRunning` field to determine deleted clusters that are running? Otherwise you'll have to recreate all the complex logic to catch all the cases.

GeekSheikh, line 1117: don't collect like this, it's dangerous. You should instead use DataSet and get back an optional value.

```scala
val test = df
  .select('timestamp)
  .as[Option[java.sql.timestamp]] // this converts it to a strongly typed dataset with an optional null
  .collect()

test.getOrElse(...)
```

src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
```diff
@@ -1098,23 +1099,50 @@ trait SilverTransforms extends SparkSessionWrapper {
...
+
+    clusterEventsBaselineFinal
+      .withColumn("state",when(col("timestamp").isin(timeStampForRemovedCluster: _*),lit("PERMANENT_DELETE")).otherwise('state))
+      .withColumn("isRunning",when(col("timestamp").isin(timeStampForRemovedCluster: _*),lit(false)).otherwise('isRunning))
+      .withColumn("unixTimeMS_state_start",when(col("timestamp").isin(timeStampForRemovedCluster: _*),lit(System.currentTimeMillis)).otherwise('unixTimeMS_state_start))
+      .withColumn("unixTimeMS_state_end",when(col("timestamp").isin(timeStampForRemovedCluster: _*),lit(System.currentTimeMillis))
```
GeekSheikh, line 1123: this logic doesn't make sense. I think you should do this differently. Move it to the end identify your billable running clusters and join on cluster_id and keep the timestamp for the deleted fields and make that the timestamp field.

GeekSheikh (changes requested):

GeekSheikh:
image

src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala

src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala

src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
```diff
@@ -1086,8 +1087,42 @@ trait SilverTransforms extends SparkSessionWrapper {
...
+    val clustersRemoved = clusterBaseDF
+      .filter('actionName.isin("permanentDelete"))
+      .select('timestamp.alias("deletion_timestamp"),'organization_id, 'cluster_id, 'userEmail.alias("deleted_by"))
+
+    val removedClusterID = clustersRemoved.select("cluster_id","deletion_timestamp").distinct()
+    val clusterEventsBaselineForRemovedCluster = clusterEventsBaseline.join(removedClusterID,Seq("cluster_id"))
 
-    clusterEventsBaseline
+    val window = Window.partitionBy('organization_id, 'cluster_id).orderBy('timestamp.desc)
+    val stateBeforeRemoval = clusterEventsBaselineForRemovedCluster.withColumn("rnk",rank().over(window)).filter('rnk === 1).drop("rnk")
```
GeekSheikh, line 1107: add row_number filter too just to be safe. Shouldn't need it but it's best practice in case there are multiple records by key.

src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
```diff
@@ -1086,8 +1087,42 @@ trait SilverTransforms extends SparkSessionWrapper {
...
+      .withColumn("isRunning",lit(false))
+      .withColumn("state",lit("PERMENANT_DELETE"))
+      .withColumn("current_num_workers",lit(0))
+      .withColumn("target_num_workers",lit(0))
+      .withColumn("unixTimeMS_state_start",'timestamp)
+      .withColumn("unixTimeMS_state_end",'timeStamp)
+      .drop("deletion_timestamp")
+    val columns: Array[String] = clusterEventsBaseline.columns
+
+    val stateDuringRemovalFinal = stateBeforeRemoval.withColumn("unixTimeMS_state_end",'deletion_timestamp).drop("deletion_timestamp").union(stateDuringRemoval).select(columns.map(col): _*)
```
GeekSheikh, line 1120: `unixTimeMS_state_end` should be `deletion_timestamp - 1` have the previous state end at the milisecond before the delete that way it follows the existing convention. Goal is illustrated in image attached

image

src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
```diff
@@ -1086,8 +1087,42 @@ trait SilverTransforms extends SparkSessionWrapper {
...
+
+    val clusterEventsBaselineFinal = clusterEventsBaseline.join(stateDuringRemovalFinal,Seq("cluster_id","timestamp"),"anti").select(columns.map(col): _*).union(stateDuringRemovalFinal)
```
GeekSheikh, line 1122: I'm seeing what appears to be some duplicate records introduced here. Please review closely and make sure the data is accurate.

image

src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
```diff
@@ -1086,8 +1087,42 @@ trait SilverTransforms extends SparkSessionWrapper {
...
+      .select('timestamp.alias("deletion_timestamp"),'organization_id, 'cluster_id, 'userEmail.alias("deleted_by"))
+
+    val removedClusterID = clustersRemoved.select("cluster_id","deletion_timestamp").distinct()
+    val clusterEventsBaselineForRemovedCluster = clusterEventsBaseline.join(removedClusterID,Seq("cluster_id"))
 
-    clusterEventsBaseline
+    val window = Window.partitionBy('organization_id, 'cluster_id).orderBy('timestamp.desc)
+    val stateBeforeRemoval = clusterEventsBaselineForRemovedCluster.withColumn("rnk",rank().over(window)).filter('rnk === 1).drop("rnk")
+
+    val stateDuringRemoval = stateBeforeRemoval
```
GeekSheikh, line 1109: Why is the permanent delete OW_runID and Pipeline_SnapTS the same as the one above it. We should be filling those. Look at all the columns and be sure to define each column appropriately.

image

GeekSheikh (changes requested):

src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
souravbaner-da: Done

src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
souravbaner-da: Done

src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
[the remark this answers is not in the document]
souravbaner-da: Done

src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
souravbaner-da: I think it is showing correct value. Can we discuss the same in our daily call

src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
souravbaner-da: Done

src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
```diff
@@ -1086,8 +1087,42 @@ trait SilverTransforms extends SparkSessionWrapper {
...
+      .withColumn("unixTimeMS_state_start", 'timestamp)
+      .withColumn("unixTimeMS_state_end", coalesce( // if state end open, use pipelineSnapTime, will be merged when state end is received
+        lead('timestamp, 1).over(stateUnboundW) - lit(1), // subtract 1 millis
+        lit(untilTime.asUnixTimeMilli)
+      ))
+
+    // Change for PR -934
+    // Get the ClusterID that has been Permenantly_Deleted
+    val clusterBaseDF = clusterBase(auditLogDF)
+    val clustersRemoved = clusterBaseDF
```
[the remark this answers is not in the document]
souravbaner-da: Not needed as we are already using clusterBase() function

PR #935, opened 2023-05-16
Null spark executor issue resolved

### LINKED ISSUE #903: [BUG] - Spark Executor table Empty
Multi-workspace deployments seem to have some issue pushing the case sensitivity setting into the thread-local sparkSession.

This can be seen in #819 

Additionally, the Spark Executor table is empty -- initial analysis points to the Timestamp field not being correctly cased resulting in null Timestamps resulting in no data making it to the targets.

sriram251-code:
closes #903

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala
@@ -676,6 +676,8 @@ trait BronzeTransforms extends SparkSessionWrapper {
     SchemaScrubber(exceptions = Array(propertiesScrubException))
   }
 
+  private def sparkCaseSensitiveValueDebug(): String = spark.conf.getOption("spark.sql.caseSensitive").getOrElse("unset")
+
   def generateEventLogsDF(database: Database,
                           badRecordsPath: String,
                           processedLogFilesTracker: PipelineTable,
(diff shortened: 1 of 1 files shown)


souravbaner-da (changes requested):
@Sriram-databricks    Please remove unnecessary debug print statement

souravbaner-da (approved):

src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala
```diff
@@ -778,6 +781,8 @@ trait BronzeTransforms extends SparkSessionWrapper {
           }
         }
 
+       val baseEventsDF = baseEventsDFRaw.withColumnRenamed("DerivedTimestamp","Timestamp")
+         logger.log(Level.INFO, s"SPARK CASE SENSITIVITY - POS1: THREAD: ${Thread.currentThread()} --> ${sparkCaseSensitiveValueDebug()}")
```
GeekSheikh, line 785: these 4 POS logs were just for debugging, let's remove these.

src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala
```diff
@@ -761,8 +763,9 @@ trait BronzeTransforms extends SparkSessionWrapper {
           } else col("Timestamp")
 
           baseDF
-            .withColumn("Timestamp", fixDupTimestamps)
+            .withColumn("DerivedTimestamp", fixDupTimestamps)
```
GeekSheikh, line 766: This is a great idea. I'm thinking we might should do something similar for all the complex field derivations. The stageId derivation could be getting null for a similar reason.

What I mean by something similar is give them a temporary column name and then rename them, drop the originals and rename it back to the target name.

* executorIdOverride
* progressCol
* stageIDColumnOverride
* stageAttemptIDColumnOverride

GeekSheikh (changes requested):

sriram251-code:
duplicate of https://github.com/databrickslabs/overwatch/pull/940.
I have moved the review comments to 940 PR.

Closing it.

PR #940, opened 2023-05-23
 Spark Executor table Empty resolved

### LINKED ISSUE #903: [BUG] - Spark Executor table Empty

sriram251-code:
closes #903


GeekSheikh (approved):
looks like this should work. Please validate at customer as described internally and once verified there, will merge.

PR #944, opened 2023-05-30
Ambiguous API call error msg fixed

### LINKED ISSUE #911: [BUG] - Several non-API errors are reporting API CALL Failed
To my recollection, module 1004 audit_log_bronze never hits an API. If this is accurate, why are we getting an API Call Failed error in the STDOUT? Please review the exception logic -- I feel like I'm seeing this error message erroneously in several places now, please double check.

image

sriram251-code:
closes #911

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ name := "overwatch"
 
 organization := "com.databricks.labs"
 
-version := "0.7.2.0"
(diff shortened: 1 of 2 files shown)


aman-db (approved):
Looks good to me

GeekSheikh (approved):

PR #953, opened 2023-06-05
948 job_dim_overlap

### LINKED ISSUE #948: [BUG] - job_status_silver -- Overlap required -- incrementalDF Lag days
job status silver -- needs to have an overlap of one day and be converted to merge insert only

Assume the following scenario

job created on 5/29 at 8:13 AM
Overwatch runs on 5/29 at 8:20 with an untilTime of 8:14 AM
The create event didn't make it to the audit events bronze table by the time the silver pipeline executed
job_status_silver will not see the create event that occurred at 8:13 because it hadn't arrived yet
the subsequent run on 5/30 does pick up the create event in audit_log_bronze but since job_status_silver's 5/30 run will have a fromTime of 8:14, the create event will never make it to silver.

Fix:
add lag of 1 day and convert job_status_silver from append to merge, insertOnly

GeekSheikh:
closes #948


brij-raghuwanshi-db (commented):
Looks Good to go.

PR #955, opened 2023-06-08
[Bug] - Added filter for response Code 200

### LINKED ISSUE #901: [BUG] -- Cluster Spec -- Node types occasionally inaccurate
We see occasional examples of clusters where the node type has been edited to a different type but the updated values are not being reflected in cluster_spec_silver / cluster_gold. This has been identified so far only in 0713, is there an issue or was this example a 1-off? Either way, what is the cause and how to protect against it.

@gueniai -- you found an example internally in demo I believe where this happened. Could you paste the query to show it and also paste a screenshot of the actual current node type?

souravbaner-da:
Added Condition for 200 Status code in SilverTransform.Scala.

Close #901

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
@@ -324,7 +324,6 @@ trait SilverTransforms extends SparkSessionWrapper {
   //  Tested in AWS, azure account logic may be slightly different
   protected def accountLogins()(df: DataFrame): DataFrame = {
     val prunedDF = df
-      .filter(responseSuccessFilter)
       .select(
         'timestamp,
         'organization_id,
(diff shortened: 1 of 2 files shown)


src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
```diff
@@ -324,6 +324,7 @@ trait SilverTransforms extends SparkSessionWrapper {
   //  Tested in AWS, azure account logic may be slightly different
   protected def accountLogins()(df: DataFrame): DataFrame = {
     val prunedDF = df
+      .filter(responseSuccessFilter)
```
GeekSheikh, line 327: remove this

src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
```diff
@@ -395,6 +396,7 @@ trait SilverTransforms extends SparkSessionWrapper {
       .rowsBetween(Window.currentRow, Window.unboundedFollowing)
     val modifiedAccountsDF = df.filter('serviceName === "accounts" &&
       'actionName.isin("add", "addPrincipalToGroup", "removePrincipalFromGroup", "setAdmin", "updateUser", "delete"))
+      .filter(responseSuccessFilter)
```
GeekSheikh, line 399: remove this

GeekSheikh (changes requested):
super minor changes -- philosophically I think we want to keep the unsuccessful account login and account mod attempts in gold. Failed account mods and logins are very important within the context of these tables. I don't want customers to have to go back to bronze to find these.

src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala

src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala

GeekSheikh (approved):

PR #963, opened 2023-06-19
response 200 backed out for jobRuns

### LINKED ISSUE #901: [BUG] -- Cluster Spec -- Node types occasionally inaccurate

GeekSheikh:
closes #901


souravbaner-da (approved):

souravbaner-da (changes requested):
Should we not add filter Condition should be added in  `protected def dbJobRunsSummary`

souravbaner-da (approved):

souravbaner-da (approved):

PR #966, opened 2023-06-20
backport getOrgId fix

GeekSheikh:
backporting critical fix applied in 0720

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/pipeline/Initializer.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/pipeline/Initializer.scala
@@ -600,7 +600,7 @@ object Initializer extends SparkSessionWrapper {
     logger.log(Level.INFO, "Initializing Config")
     val config = new Config()
     if(organizationID.isEmpty) {
-      config.setOrganizationId(getOrgId)
+      config.setOrganizationId(getOrgId(apiUrl))
     }else{ // is multiWorkspace deployment since orgID is passed
       logger.log(Level.INFO, "Setting multiworkspace deployment")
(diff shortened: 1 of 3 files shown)


src/main/scala/com/databricks/labs/overwatch/pipeline/Initializer.scala
```diff
@@ -561,10 +561,38 @@ object Initializer extends SparkSessionWrapper {
...
+  def getOrgId: String = getOrgId(None)
+
+  /**
+   * Returns the local workspace orgID, or the orgID contained in the input apiURL
+   * When getOrgId is called from a Future the apiURL MUST be passed to prevent failure to get orgID
+   * since dbutils.notebook.getContext.apiUrl.get returns None and None.get fails
+   * @param apiUrl apiURL of the workspace to get the orgId
+   * @return
+   */
+  def getOrgId(apiUrlInput: Option[String]): String = {
```
gueniai, line 587: We should also call this one from `Initializer.initConfigState`

gueniai: also from `Helpers.getWorkspaceByDatabase`

gueniai (changes requested):
There are 2 other places where this function is called that we should backport as well.

src/main/scala/com/databricks/labs/overwatch/pipeline/Initializer.scala
GeekSheikh: nice catch, thank you!

gueniai (approved):
ship it!

PR #969, opened 2023-06-20
968 - dup shardName Field Handled

### LINKED ISSUE #968: [BUG] - PVC - Some PVC customers have shardName in audit root and in requestParams
the root field `shardName` from auditMasterSchema must be removed if present to ensure no duplicates from root and requestParams nested.

GeekSheikh:
closes #968


gueniai (approved):
LGTM

PR #970, opened 2023-06-21
refactored database write function

### LINKED ISSUE #702: [Enhancement] : Refactor Database.scala write method
Scenario  :Now we have a single write method which is being used for all types of write ex( Streaming write, write which has merge condition, generic write). As the single write function is implementing different kind of writes and all the writes have different types of arguments which has been used to perform the write, its very complex to make changes and refactor the write method.

Solution: We can have a Trait/Abstract class which will contain a write method and we can have different types of writer implementing/extending the trait.

aman-db:
closes #702

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/env/Database.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/env/Database.scala
@@ -341,8 +341,13 @@ class Database(config: Config) extends SparkSessionWrapper {
    *                           concurrency locking.
    * @return
    */
-  def write(df: DataFrame, target: PipelineTable, pipelineSnapTime: Column, maxMergeScanDates: Array[String] = Array(),
-            preWritesPerformed: Boolean = false): Unit = {
+  def write(
+             df: DataFrame,
+             target: PipelineTable,
+             pipelineSnapTime: Column,
+             maxMergeScanDates: Array[String] = Array(),
+             preWritesPerformed: Boolean = false
+           ): Unit = {
 
     // append metadata to source DF and cleanse as necessary
     val finalDF = if (!preWritesPerformed) {
@@ -460,7 +465,7 @@ class Database(config: Config) extends SparkSessionWrapper {
    * @param daysToProcess
    * @return
    */
-  def ifCachedNeeded(target: PipelineTable, daysToProcess: Option[Int] = None): Boolean = {
+  def preWriteCacheRequired(target: PipelineTable, daysToProcess: Option[Int] = None): Boolean = {
     daysToProcess.getOrElse(1000) < 5 &&   // this can be inside a function
       !target.autoOptimize &&
       !SparkSessionWrapper.parSessionsOn &&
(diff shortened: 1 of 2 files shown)


src/main/scala/com/databricks/labs/overwatch/env/Database.scala
```diff
@@ -248,82 +341,25 @@ class Database(config: Config) extends SparkSessionWrapper {
...
-          case e: Throwable =>
-            val fullMsg = PipelineFunctions.appendStackStrace(e, "Exception in writeMethod:")
-            logger.log(Level.ERROR, s"""Exception while writing to ${target.tableFullName}"""" + Thread.currentThread().getName + " Error msg:" + fullMsg)
-            throw e
-        }
-      }
-      logger.log(Level.INFO, s"Completed write to ${target.tableFullName}")
+    target.writeMode match {
+      case WriteMode.merge => batchWriter(finalDF, target, maxMergeScanDates)
+      case _ => streamWriter(finalDF, target)
```
GeekSheikh, line 362: uhhh soo what about WriteMode.append and overwrite? Below's a screenshot of what used to handle it.

You have the following conditions: 
* merge
* Not Merge -- this can either be
  * Stream
  * Append/Overwrite

image

GeekSheikh (changes requested):
you forgot a really big part of the database writer. ;)

src/main/scala/com/databricks/labs/overwatch/env/Database.scala
aman-db: @GeekSheikh This else block is covered in streamWriter function.

src/main/scala/com/databricks/labs/overwatch/env/Database.scala
```diff
@@ -232,6 +232,99 @@ class Database(config: Config) extends SparkSessionWrapper {
...
+    val mergeCondition = getMergeCondition(df, target, maxMergeScanDates)
+
+    val mergeDetailMsg =
+      s"""
+         |Beginning upsert to ${target.tableFullName}.
+         |MERGE CONDITION: $mergeCondition
+         |""".stripMargin
+    logger.log(Level.INFO, mergeDetailMsg)
+    spark.conf.set("spark.databricks.delta.commitInfo.userMetadata", config.runID)
+    // TODO -- when DBR 9.1 LTS GA, use LSM (low-shuffle-merge) to improve pipeline
```
gueniai, line 274: LSM is now enabled by default as of 10.4 :)

gueniai (approved):
Left you one super minor nit-pick for a comment that can be deprecated, but otherwise this looks good to me!

src/main/scala/com/databricks/labs/overwatch/env/Database.scala
```diff
@@ -232,6 +232,99 @@ class Database(config: Config) extends SparkSessionWrapper {
...
+
+    spark.conf.unset("spark.databricks.delta.commitInfo.userMetadata")
+  }
+
+  /**
+   * Function to write the streaming data
+   * @param df
+   * @param target
+   */
+  def streamWriter(df: DataFrame, target: PipelineTable): Unit = {
```
gueniai, line 286: Since this function does both stream and batchWrites, I'd switch it to dataframeWriter or something else. And change batchWriter to mergeWriter or something else

gueniai (changes requested):
minor readability change request

src/main/scala/com/databricks/labs/overwatch/env/Database.scala
aman-db: done

gueniai (approved):

PR #983, opened 2023-06-28
906 update instance details

### LINKED ISSUE #904: instanceDetails and dbuCostDetails should have verifyMinSchema
complete with #906 

to ensure no future issues on types from the CSV resources intanceDetails and dbuCostDetails should have a verifyMinimumSchema method applied to the DF before being written. 

The implicit schema definition can see memory gb 33 as an integer and a later update may have another node type with 24.2 requiring the field to be a double resulting in a schema error.

verifyMinSchema will protect against this in future releases.

### LINKED ISSUE #906: [FEAT] - instanceDetails -> updateInstanceDetails
Complete with #904 

when new version have new node types in instanceDetails we should be able to let the customer update it.

Create `Helpers.updateInstanceDetails()`

The function should insert new node types and update the hourly_dbus for existing types.

GeekSheikh:
closes #906 
closes #904 

Offers a way for customers to easily update their instanceDetails table to the latest version.

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/utils/RefreshNodeDetails.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/utils/RefreshNodeDetails.scala
@@ -168,13 +168,20 @@ private class RefreshNodeDetails(
 
     // full join to get original df with all values from git where git fields (less keys) are renamed to be
     // suffixed with _update
-    val mergeUpdatesDF = instanceDetailsDF
-      .join(latestInstanceDetailsFromGitDF, keyFields, "full")
+    val mergeUpdatesDF = latestInstanceDetailsFromGitDF
+      .join(
+        instanceDetailsDF
+          .withColumnRenamed("organization_id", "target_organization_id")
+          .withColumnRenamed("API_Name", "target_API_Name"),
+        'target_organization_id === 'organization_id &&
+          'target_API_Name === 'API_Name
+        , "left")
 
     // identify record status as determined by whether it's a new key, changed value, or no change
     val isUpdateLogic = 'Hourly_DBUs =!= 'Hourly_DBUs_update && coalesce('Hourly_DBUs, 'Hourly_DBUs_update).isNotNull
+    val isInsertLogic = 'target_organization_id.isNull && 'target_API_Name.isNull // not matched - insert
     val recordStatus = when(isUpdateLogic, array(lit("update"), lit("newActive")))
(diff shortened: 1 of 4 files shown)


src/main/scala/com/databricks/labs/overwatch/utils/RefreshNodeDetails.scala
```diff
@@ -0,0 +1,344 @@
...
+    // append _update to all fields except the key fields
+    val latestInstanceDetailsFromGitDF = fieldsToUpdate
+      .foldLeft(getLatestInstanceDetailsDF)((df, f) =>
+        df.withColumnRenamed(f, s"${f}_update")
+      )
+
+    // full join to get original df with all values from git where git fields (less keys) are renamed to be
+    // suffixed with _update
+    val mergeUpdatesDF = instanceDetailsDF
+      .join(latestInstanceDetailsFromGitDF, keyFields, "full")
```
aman-db, line 172: we should be using left join , instead of full join for a better performance. Also we are not making any changes in the target table, so doing a left join will be more efficient

src/main/scala/com/databricks/labs/overwatch/utils/RefreshNodeDetails.scala
```diff
@@ -0,0 +1,344 @@
...
+
+    // identify record status as determined by whether it's a new key, changed value, or no change
+    val isUpdateLogic = 'Hourly_DBUs =!= 'Hourly_DBUs_update && coalesce('Hourly_DBUs, 'Hourly_DBUs_update).isNotNull
+    val recordStatus = when(isUpdateLogic, array(lit("update"), lit("newActive")))
+      .when('Hourly_DBUs.isNull, array(lit("insert")))
+      .otherwise(array(lit("noUpdate")))
+
```
aman-db, line 179: we should be using joining keys to identify updates and inserts.

GeekSheikh: agree for inserts -- changed. For updates though, I don't want to update the full record just because the keys match, I only want to update if the keys match AND the hourly_dbus changed. I want to disregard all other record changes as the instance types should be immutable.

src/main/scala/com/databricks/labs/overwatch/utils/RefreshNodeDetails.scala
GeekSheikh: done, thanks.

aman-db (approved):
Looking Good now :)

PR #1002, opened 2023-07-25
Add Verify Min Schema in audit_log_bronze

### LINKED ISSUE #974: [BUG] - AuditLogBronze (1004) does not enforce Min Schema
When writing the audit_log_bronze -- there's no verifyMinSchema to ensure the output data is commensurate with the required output format.

Example `requestParams.enable_elastic_disk` -- if this is all null in the incremental audit data (no cluster create / edit) then it will be null for all records and will not be implicitly cast to a boolean (as necessary); thus an error will occur when the OW tries to merge string of all nulls into boolean.

This was observed at PVC customer with parquet formatted audit logs but I believe it would be best practice to ensure the land schema is always enforced.

**Solution**
At the end of the code snippet below, add `.verifyMinimumSchema(Schema.auditMasterSchema)` to ensure all fields are forced to land in the right type.

https://github.com/databrickslabs/overwatch/blob/736ec19767159f5204e78dc00b5c803259eae806/src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala#L407-L417

souravbaner-da:
Add Verify Min Schema in audit_log_bronze
Close #974

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ name := "overwatch"
 
 organization := "com.databricks.labs"
 
-version := "0.7.2.0.2"
+version := "0.7.2.1"
 
 scalaVersion := "2.12.12"
 scalacOptions ++= Seq("-Xmax-classfile-name", "78")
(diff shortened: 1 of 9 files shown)


sriram251-code (changes requested):
please test it in AWS,Azure and GCP

sriram251-code (approved):

PR #1028, opened 2023-08-29
fix cluster log

### LINKED ISSUE #1021: Remote cluster logs are not getting captured for some remote workspace
Issue: In MWS deployment for some of the remote workspace the cluster log are not getting captured.

Findings:
For remote workspace when we are calling mount point api we are getting "" as a mount point and we are getting a source which is for the "" mount point.
Hence when we are deriving the cluster log location the source of the "" is getting appended in all the cluster log path. Hence making the path corrupted.

Fix:
Add filter .filter(col("mount_point") =!= "") while getting the mount point to source mapping.

sriram251-code:
closes #1021

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala
@@ -415,6 +415,7 @@ trait BronzeTransforms extends SparkSessionWrapper {
           .withColumn("date",
             split(expr("filter(filenameAR, x -> x like ('date=%'))")(0), "=")(1).cast("date"))
           .drop("filenameAR")
+          .verifyMinimumSchema(Schema.auditMasterSchema)
       } else {
         throw new Exception(auditLogsFailureMsg)
       }
--- a/src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala
@@ -1023,26 +1023,42 @@ trait BronzeTransforms extends SparkSessionWrapper {
     // This captures clusters that have not been edited/restarted since the last run with
     // log confs as they will not be in the audit logs
     val latestSnapW = Window.partitionBy('organization_id, 'cluster_id).orderBy('Pipeline_SnapTS.desc)
-    val newLogDirsNotIdentifiedInAudit = clusterSnapshot
-      .join(incrementalClusterIDs, Seq("cluster_id"))
-      .withColumn("snapRnk", rank.over(latestSnapW))
-      .filter('snapRnk === 1)
-      .withColumn("cluster_log_conf",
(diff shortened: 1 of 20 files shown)


src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala
```diff
@@ -894,7 +894,7 @@ trait BronzeTransforms extends SparkSessionWrapper {
   private[overwatch] def getAllEventLogPrefix(inputDataframe: DataFrame, apiEnv: ApiEnv): DataFrame = {
     try{
       val mountMap = getMountPointMapping(apiEnv) //Getting the mount info from api and cleaning the data
-        .filter(col("mount_point") =!= "/")
+        .filter(col("mount_point") =!= "/").filter(col("mount_point") =!= "")
```
aman-db, line 897: @sriram251-code - should we apply the trim function on the mount_point column before doing this type of comaparision?

src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala
```diff
@@ -1024,26 +1023,42 @@ trait BronzeTransforms extends SparkSessionWrapper {
...
-      .filter('cluster_id.isNotNull && 'cluster_log_conf.isNotNull)
-      .select('cluster_id, 'cluster_log_conf)
+
+    val newLogDirsNotIdentifiedInAudit = {
+      if(clusterSnapshot.isEmpty){
+        spark.emptyDataFrame
+      }else{
+        clusterSnapshot
+          .join(incrementalClusterIDs, Seq("cluster_id"))
+          .withColumn("snapRnk", rank.over(latestSnapW))
```
aman-db, line 1033: @sriram251-code we should use row_number instead of rank, just we avoid a condition where we have multiple rows having rank = 1

sriram251-code: Point well taken but the code was previously present i have just added some exception handling. Moving to row_number might need some regression testing that we can achieve after our validation module.

aman-db: okay

src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala
```diff
@@ -894,7 +894,7 @@ trait BronzeTransforms extends SparkSessionWrapper {
   private[overwatch] def getAllEventLogPrefix(inputDataframe: DataFrame, apiEnv: ApiEnv): DataFrame = {
     try{
       val mountMap = getMountPointMapping(apiEnv) //Getting the mount info from api and cleaning the data
-        .filter(col("mount_point") =!= "/")
+        .filter(col("mount_point") =!= "/").filter(trim(col("mount_point")) =!= "") //Added for #1021
```
souravbaner-da, line 897: @sriram251-code  Instead of 2 filter condition can we put both as and condition. and also if you are adding trim in one filter condition please add in both.

souravbaner-da (approved):

gueniai (approved):

PR #1032, opened 2023-09-05
1031_bug_fix_empty_cluster_policy_API_response

### LINKED ISSUE #1031: [BUG] Module 1008: Cluster Policies, getting failed if there are no cluster policies in the workspace
Module 1008 Cluster Policies is failing if there are no cluster policies defined in the workspace, ideally for an empty response an empty dataframe should be returned from APIV2.execute.asDF method but it is throwing an exception.

aman-db:
closes #1031


gueniai (approved):
I'm always happier when I see a unit test in a PR 🙌

PR #1065, opened 2023-10-19
Hotfix for Audit_log_bronze and NotebookCommands_gold

### LINKED ISSUE #1066: [Bug] - Getting null notebook_id in notebookCommands_gold
Customer is getting null notebook_id in notebookCommands_gold.

### LINKED ISSUE #1067: [Bug] - auditLog_Bronze is failing when extra columns are being sent from event hub.
Currently we are taking all the field from properties schema from eventHub. But due to some configuration if we are getting some extra field in this schema then audit log bronze is failing.

souravbaner-da:
Close #1066 
Close #1067

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala
@@ -339,7 +339,7 @@ trait BronzeTransforms extends SparkSessionWrapper {
       val azureAuditSourceFilters = 'Overwatch_RunID === lit(overwatchRunID) && 'organization_id === organizationId
       val rawBodyLookup = auditRawLand.asDF
         .filter(azureAuditSourceFilters)
-      println(s"Columns in rawBodyLookup are ${rawBodyLookup.toDF().columns.mkString("Array(", ", ", ")")}")
+
       val schemaBuilders = auditRawLand.asDF
         .filter(azureAuditSourceFilters)
         .withColumn("parsedBody", structFromJson(spark, rawBodyLookup, "deserializedBody"))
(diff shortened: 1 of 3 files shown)


src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala
```diff
@@ -349,8 +350,18 @@ trait BronzeTransforms extends SparkSessionWrapper {
...
+          "properties.actionName",
+          "properties.logId",
+          "properties.requestId",
+          "properties.requestParams",
+          "properties.response",
+          "properties.serviceName",
+          "properties.sessionId",
+          "properties.sourceIPAddress",
+          "properties.userAgent")
+        .drop("properties")
```
aman-db, line 364: Can we add these column list in Schema.scala and then use this list instead of passing the hardcoded column names ?

src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala
```diff
@@ -363,15 +374,28 @@ trait BronzeTransforms extends SparkSessionWrapper {
...
```
aman-db, line 388: Can we add these column list in Schema.scala and then use this list instead of passing the hardcoded column names ?

src/main/scala/com/databricks/labs/overwatch/pipeline/Gold.scala
```diff
@@ -310,7 +310,7 @@ class Gold(_workspace: Workspace, _database: Database, _config: Config)
   lazy private val appendNotebookCommandsFactProcess: () => ETLDefinition = {
     () =>
       ETLDefinition(
-        BronzeTargets.auditLogsTarget.asIncrementalDF(notebookCommandsFactModule, BronzeTargets.auditLogsTarget.incrementalColumns,1),
+        BronzeTargets.auditLogsTarget.asIncrementalDF(notebookCommandsFactModule, BronzeTargets.auditLogsTarget.incrementalColumns,30),
```
gueniai, line 313: How does this affect the performance of this table? There are potentially many rows in Audit logs that would be brought in when using a 30 day lag

PR #1083, opened 2023-11-08
Added JobRun_Silver in buildClusterStateFact for Cluster E…

### LINKED ISSUE #1020: Impute automated clusters termination as a last resort
There have been a few issues with the platform where we might miss the terminating event for an automated cluster. Can we investigate if it would be possible to impute the cluster termination from the job termination timestamp?

souravbaner-da:
Added Change for imputation for clusters termination.
Close #1020

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
@@ -246,6 +246,7 @@ trait GoldTransforms extends SparkSessionWrapper {
                                        dbuCostDetailsTarget: PipelineTable,
                                        clusterSnapshot: PipelineTable,
                                        clusterSpec: PipelineTable,
+                                       jrsilverDF: DataFrame,
                                        pipelineSnapTime: TimeTypes
                                      )(clusterStateDetail: DataFrame): DataFrame = {
     val instanceDetailsDF = instanceDetailsTarget.asDF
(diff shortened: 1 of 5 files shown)


aman-db (approved):

dougspadotto:
Hi. Is there an estimate on when this will be merged?

gueniai:
@dougspadotto We're targetting first week of December

src/main/scala/com/databricks/labs/overwatch/pipeline/GoldTransforms.scala
```diff
@@ -402,8 +403,21 @@ trait GoldTransforms extends SparkSessionWrapper {
       (driverComputeCost + driverDBUCost + workerComputeCost + workerDBUCost).alias("total_cost")
     )
 
-    clusterPotential
+    val clsfDF = clusterPotential
       .select(clusterStateFactCols: _*)
+
+    val clsfDF1 =clsfDF.filter('state === "TERMINATING").filter('isAutomated === true).filter('days_in_state > 2)
```
gueniai, line 409: This handles one of the corner cases, but doesn't help us in cases where the job finishes and we're missing the TERMINATING event altogether. 
In my opinion, we should approach it from the the other side, find the job runs that used an automated cluster and terminated, then find those on the list for clsf and mark them as terminated

gueniai (changes requested):

src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
```diff
@@ -1020,13 +1023,44 @@ trait SilverTransforms extends SparkSessionWrapper {
 
     val invalidEventChain = lead('runningSwitch, 1).over(stateUnboundW).isNotNull && lead('runningSwitch, 1)
       .over(stateUnboundW) === lead('previousSwitch, 1).over(stateUnboundW)
-    val clusterEventsBaseline = clusterEventsDF
+
+
+    val clusterEventsDF1 = clusterEventsDF
```
gueniai, line 1028: I know these inner variables are hard to name but let's find something more descriptive than xyzDF1. Longer variable names are fine

gueniai (changes requested):
Minor nitpick around more declarative variable names, I think this is almost done. Did you do a second run to ensure the terminated_impute was respected as a terminal event and the cluster's events weren't progressed anymore? (I couldn't find it in the proof)

src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala

gueniai (approved):

PR #1102, opened 2023-11-29
traceability implemented

### LINKED ISSUE #891: Increase Traceability for API calls[Draft]
All the performed Api calls should register the below metrics into a table 

Request filter which contains all the request parameters.
Count of the response.

- [ ] design table schema for API trace events
- [ ] implement event accumulators in API process to accumulate and strategically persist these call events to the table defined above
- [ ] When parallel API Calls made capture the mini-batch details (i.e. in clusterEvents capture all the cluster_ids that are to be scanned for new events.
- [ ] what else?

sriram251-code:
closes #891


gueniai (approved):
LGTM, ship it!

PR #1110, opened 2023-12-06
Fix Helper Method to Instantiate Remote Workspaces

### LINKED ISSUE #1090: Fix helper method to instantiate remote workspaces
Currently `Helpers.getWorkspaceByDatabase` does not work if you are trying to instantiate a remote workspace, it may seem like it works, but it still returns the local workspace object.

souravbaner-da:
Close #1090

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ name := "overwatch"
 
 organization := "com.databricks.labs"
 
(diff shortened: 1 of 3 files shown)


aman-db (approved):
LGTM

gueniai (approved):

PR #1116, opened 2023-12-15
implemented recon

### LINKED ISSUE #1029: Data Reconciliation for Validation framework
Once we release a new version of the overwatch this module will ensure whether the data is consistent across the current release and previous release. 
In order to perform the data reconciliation, we need two overwatch deployments with current and previous versions having the same configurations(primordial date, max days) and then check whether the data between deployments is matching or not. This validation check should be configurable to include/exclude tables or even the columns specific to tables. We should also be able to configure the number of days to be considered for comparison.

After running the reconciliation it will generate a report. The report will contain for each table

Total number of rows
Total number of rows matching
Total number of rows not matching
ETC

sriram251-code:
closes #1029

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/validation/DataReconciliation.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/validation/DataReconciliation.scala
@@ -2,6 +2,7 @@ package com.databricks.labs.overwatch.validation
 
 import com.databricks.labs.overwatch.env.Workspace
 import com.databricks.labs.overwatch.pipeline._
+import com.databricks.labs.overwatch.utils.Helpers.getAllPipelineTargets
 import com.databricks.labs.overwatch.utils.{Helpers, ReconReport, SparkSessionWrapper}
 import org.apache.log4j.Logger
 import org.apache.spark.sql.DataFrame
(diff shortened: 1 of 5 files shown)


souravbaner-da (approved):

src/main/scala/com/databricks/labs/overwatch/validation/DataReconciliation.scala
```diff
@@ -0,0 +1,327 @@
...
+
+  }
+
+  /**
+   * This method fetches all targets for a workspace.
+   *
+   * @param workspace : Workspace object
+   * @return ParArray of PipelineTable
+   */
+  private[overwatch] def getAllTargets(workspace: Workspace): ParArray[PipelineTable] = {
```
gueniai, line 226: this function is heavily used during our cloning process, might be worth it to put it in Helpers so that we can use this instead

src/main/scala/com/databricks/labs/overwatch/validation/DataReconciliation.scala
```diff
@@ -0,0 +1,327 @@
...
+    (b.getAllTargets ++ s.getAllTargets ++ g.getAllTargets).filter(_.exists(dataValidation = true, catalogValidation = false)).par
+  }
+
+  /**
+   * This method performs basic reconciliation between two arrays of organization IDs.
+   *
+   * @param sourceOrgIDArr : Array of organization IDs from source ETL
+   * @param targetOrgIDArr : Array of organization IDs from target ETL
+   */
+  private[overwatch] def performBasicRecon(sourceOrgIDArr:Array[String],targetOrgIDArr:Array[String]):Unit = {
```
gueniai, line 239: minor nitpick: add spaces

gueniai (commented):
2 small comments on this one. Great job with the PR and feature in general, I particularly liked the docstrings!

src/main/scala/com/databricks/labs/overwatch/validation/DataReconciliation.scala
sriram251-code: done

gueniai (approved):

PR #1133, opened 2024-01-19
Bug fix sys table integration for sql_endpoints and migration validation

aman-db:
Adding fix for the following - 
* Validation for migration is failing for the first run
* Sql_endpoint is now not a mandatory field in configs , previously it was failing if sql_endpoints is not present in the configs


sriram251-code (approved):

gueniai (approved):
LGTM

PR #1135, opened 2024-01-19
Handle CLSD Cluster Impute when jrcp and clusterSpec is Empty

souravbaner-da:
in CLSD Impute Cluster was failing when JR and Cluster_Spec any one of them is empty. With this fix that has been handled gracefully

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
@@ -1034,9 +1034,8 @@ trait SilverTransforms extends SparkSessionWrapper {
       refinedClusterEventsDF
     }else{
       val refinedClusterEventsDFFiltered = refinedClusterEventsDF
-        .withColumn("last_state", first(col("state")).over(orderingWindow))
         .withColumn("row", row_number().over(orderingWindow))
-        .filter('last_state =!= "TERMINATING" && 'row === 1)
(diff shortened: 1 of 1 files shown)


src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
```diff
@@ -1030,32 +1030,35 @@ trait SilverTransforms extends SparkSessionWrapper {
...
 
-    val refinedClusterEventsDFFiltered = refinedClusterEventsDF
-      .withColumn("last_state", first(col("state")).over(orderingWindow))
-      .withColumn("row", row_number().over(orderingWindow))
-      .filter('last_state =!= "TERMINATING" && 'row === 1)
+    val clusterEventsFinal  = if (jrsilverDF.isEmpty || clusterSpec.asDF.isEmpty) {
+      refinedClusterEventsDF
+    }else{
+      val refinedClusterEventsDFFiltered = refinedClusterEventsDF
+        .withColumn("last_state", first(col("state")).over(orderingWindow))
```
gueniai, line 1037: I didn't catch this last time, but I think this is unnecessary, you can just get the "state" of whichever row ends up being the first in the window.

souravbaner-da: @gueniai  addressed

gueniai (approved):
LGTM!

PR #1154, opened 2024-02-08
added hotfix for warehouse_spec_silver

aman-db:
Adding fix for the schema mis match in warehouse_spec_silver from second run onwards.


gueniai (commented):
The fix looks good, but the PR is wanting to merge into main instead of 0810

gueniai (approved):

PR #1184, opened 2024-04-17
1178 missing cluster events 810

### LINKED ISSUE #1178: Missing Custer Events
On some perticular scenario we found out that some of the cluster events are missing.(perticularly the events which are generated on starting of the cluster).

Root Cause Analysis:
The cluster IDs we are getting from audit log
Ex we ran the job at 3pm 8-API-2024 we got data for clusters which has some activities from 7-apr 10pm to 8-apr 1pm

but when we perform cluster events api call we made the call from 7-apr 3pm till 8-apr 3pm hence we lost the event which has occured from 7-apr 10pm till 3pm.

Solution:
1)Simply while making cluster events api call add 1 day lag to from time while making the api call.

sriram251-code:
Closes #1178 

We have 1 day as lag in cluster events while making api calls.


sriram251-code:
Already had a PR for this

PR #1191, opened 2024-04-29
Change the No New Streaming Data Found msg for 3016

souravbaner-da:
Change the No New Streaming Data Found error msg for Module 3016


gueniai:
As discussed, no need for this change, since we already emit the "No new streaming data found" message.

PR #1194, opened 2024-05-02
Initializer calling fixed for multiworkspace

### LINKED ISSUE #1192: Optimizer job is failing
Its Observed that the Optimizer job is failing.

sriram251-code:
closes #1192 

Previously orgID,ApiURL and Cloud all are derived automatically.
Now we are getting the API URL from config table and OrgID from pipeline report.
Now while calling the Initializer function from Optimizer we are passing the OrgID and Api URL .

### DIFF
--- a/src/main/scala/com/databricks/labs/overwatch/Optimizer.scala
+++ b/src/main/scala/com/databricks/labs/overwatch/Optimizer.scala
@@ -71,24 +71,22 @@ object Optimizer extends SparkSessionWrapper{
    * @return workspace object
    */
   private def getLatestWorkspace(overwatchETLDB: String,orgId: String): Workspace = {
-   val latestSuccessState = getLatestSuccessState(overwatchETLDB,orgId).limit(1)
-   val orgID = latestSuccessState.select("organization_id").collect().map(x => x(0).toString)
-    val params = latestSuccessState
-      .selectExpr("inputConfig.*")
-      .as[OverwatchParams]
+
+     val ( orgID: Option[String], params: OverwatchParams) = getLatestSuccessState(overwatchETLDB,orgId)
+      .select("organization_id", "inputConfig")
+      .as[(Option[String], OverwatchParams)]
       .first
 
-    val args = JsonUtils.objToJson(params).compactString
-    val prettyArgs = JsonUtils.objToJson(params).prettyString
-    logger.log(Level.INFO, s"ARGS: Identified config string is: \n$prettyArgs")
+    val args = JsonUtils.objToJson(params)
+    logger.log(Level.INFO, s"ARGS: Identified config string is: \n${args.prettyString}")
     try{
-      val apiURL = getApiURL(orgID(0),params.dataTarget.get)
+      val apiURL = getApiURL(orgID.get,params.dataTarget.get)
       logger.log(Level.INFO,"Running optimization for Multiworkspace deployment")
-      Initializer(args,apiURL=Some(apiURL), organizationID = Some(orgID(0)))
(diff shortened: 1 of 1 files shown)


aman-db (approved):

src/main/scala/com/databricks/labs/overwatch/Optimizer.scala
```diff
@@ -31,22 +31,68 @@ object Optimizer extends SparkSessionWrapper{
...
+
   /**
    * Derive the workspace from the config supplied in the last run
    * @param overwatchETLDB name of the Overwtch ETL database
    * @return workspace object
    */
   private def getLatestWorkspace(overwatchETLDB: String,orgId: String): Workspace = {
-    val params = getLatestSuccessState(overwatchETLDB,orgId)
+   val latestSuccessState = getLatestSuccessState(overwatchETLDB,orgId).limit(1)
+   val orgID = latestSuccessState.select("organization_id").collect().map(x => x(0).toString)
```
neilbest-db, line 75: . . . then here you could simply say:
```scala
 val ( orgID: Option[String], params: OverwatchParams) = latestSuccessState
```
or even combine the two statements if you don't need the `latestSuccessState` tuple anywhere else.

I did this as a test to make sure I was giving you good advice:

image

src/main/scala/com/databricks/labs/overwatch/Optimizer.scala
```diff
@@ -31,22 +31,68 @@ object Optimizer extends SparkSessionWrapper{
...
+    val params = latestSuccessState
       .selectExpr("inputConfig.*")
       .as[OverwatchParams]
       .first
 
     val args = JsonUtils.objToJson(params).compactString
```
neilbest-db, line 81: these two calls to `JsonUtils.objToJson` seem redundant.  consider instead:

```scala
val args = JsonUtils.objToJson(params)
logger.log(Level.INFO, s"ARGS: Identified config string is: \n${args.prettyString}")
try {
  val apiURL = getApiURL( orgID(0), params.dataTarget)
```
. . .
```scala
  Initializer( args.compactString, apiURL, orgID)
```
. . . 

Then you already have the `Option` types that you need, right?

neilbest-db (commented):
I made some suggestions to make your code cleaner.  Let Scala do the work!

src/main/scala/com/databricks/labs/overwatch/Optimizer.scala
```diff
@@ -31,22 +31,68 @@ object Optimizer extends SparkSessionWrapper{
...
+
+
   /**
    * Derive the workspace from the config supplied in the last run
    * @param overwatchETLDB name of the Overwtch ETL database
    * @return workspace object
    */
   private def getLatestWorkspace(overwatchETLDB: String,orgId: String): Workspace = {
-    val params = getLatestSuccessState(overwatchETLDB,orgId)
+   val latestSuccessState = getLatestSuccessState(overwatchETLDB,orgId).limit(1)
```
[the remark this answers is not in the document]
sriram251-code: good point implemented

src/main/scala/com/databricks/labs/overwatch/Optimizer.scala
sriram251-code: implemented

src/main/scala/com/databricks/labs/overwatch/Optimizer.scala
sriram251-code: implemented

gueniai (approved):

PR #1213, opened 2024-05-15
Convert spec.custom_tags from struct type to map type

souravbaner-da:
Changed the spec.custom_tags from struct type to map type


neilbest-db (approved):
Looks good to me.  Much simpler than I expected.  Does this scenario call for creating/modifying a minimum schema, or would a minimum schema simplify this or make it more consistent with other similar scenarios?  (I don't know the answer — just brainstorming.). I can study it some more tomorrow, but if you are certain that this fixes the problem then let's move forward.  We can always look at it again another time.

PR #1224, opened 2024-05-23
Show `DataFrame` records in logs

neilbest-db:
See doc string in `DataFrameSyntax.scala`:

https://github.com/databrickslabs/overwatch/blob/e5dd9882ce0705b835aea3cfd147b09d690b3830/src/main/scala/com/databricks/labs/overwatch/utils/DataFrameSyntax.scala#L13-L99


src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
```diff
@@ -1313,6 +1313,15 @@ trait SilverTransforms extends SparkSessionWrapper {
     // eagerly force this highly reused DF into cache()
     jobRunsLag30D.count()
 
+    
+    // TODO: remove or comment out or change log level or . . .
+
+    logger.log( Level.INFO, "Showing first 5 rows of `jobRunsLag30D`:")
```
sriram251-code, line 1319: Why we r printing  5 rows of the data in logs?

src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
```diff
@@ -1313,6 +1313,15 @@ trait SilverTransforms extends SparkSessionWrapper {
...
+
+    jobRunsLag30D
+      .showLines(5, 20, true)
+      .foreach( logger.log( Level.INFO, _))
+
```
sriram251-code, line 1324: why we r printing it?

sriram251-code (changes requested):

src/main/scala/com/databricks/labs/overwatch/utils/DataFrameSyntax.scala
```diff
@@ -0,0 +1,69 @@
...
+    def showLines( numRows: Int): Iterator[String] =
+      showLines( numRows, truncate = true)
+
+    def showLines( truncate: Boolean): Iterator[String] =
+      showLines( 20, truncate)
+
+    def showLines( numRows: Int, truncate: Boolean): Iterator[String] =
+      if( truncate) {
+        showLines( numRows, truncate = 20)
+      } else {
```
souravbaner-da, line 51: @neilbest-db  why 20 count here?

src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
neilbest-db: It's just to demonstrate a temporary debug output.  Any developer can use `.showLines()` anywhere & anyhow they want to during development and then remove it, comment it out, or conditionally report something at a higher log level — whatever.

src/main/scala/com/databricks/labs/overwatch/pipeline/SilverTransforms.scala
neilbest-db: It provides an alternative to copy-pasting so much code into notebooks by providing feedback in the logs.  Any transformation is available at that point, naturally.

src/main/scala/com/databricks/labs/overwatch/utils/DataFrameSyntax.scala
neilbest-db: Like the `truncate` argument for [native](https://spark.apache.org/docs/3.3.0/api/scala/org/apache/spark/sql/Dataset.html#show(numRows:Int,truncate:Int,vertical:Boolean):Unit) `.show()` method it limits the column width of the formatted output.

Also note [this](https://github.com/databrickslabs/overwatch/issues/1227#issuecomment-2135408925):
> I have overloaded `.showLines()` with the same set of method signatures as Spark's `Dataset.show()` for API consistency.

neilbest-db:
Pasted docstring into initial comment above for reference.

neilbest-db:
@sriram251-code, do you need more info before we move ahead with this?  I don't think it's right for me to mark the conversations resolved . . . are you satisfied or still have doubts?

sriram251-code:
I am fine with the comments

sriram251-code (approved):

neilbest-db:
closes #1227

gueniai (changes requested):
This needs to be controlled by a config (an Overwatch scoped spark config will do) to turn this logging on/off, with a default being false

neilbest-db:
> This needs to be controlled by a config (an Overwatch scoped spark config will do) to turn this logging on/off, with a default being false

Done.  Please review.

gueniai (approved):

PR #1239, opened 2024-06-18
append null columns from cluster snapshot for cluster_spec_silver

### LINKED ISSUE #1000: Some clusters data is incomplete in cluster_gold
In our internal deployment, `num_workers` and `driver_node_type` are NULL for a few clusters (interactive, not new, and active) We should review

sriram251-code:
closes #1000 

We observed few columns ware missing in cluster spec silver. For those columns we are building the data from cluster_snapshot.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ name := "overwatch"
 
 organization := "com.databricks.labs"
 
-version := "0.8.1.1"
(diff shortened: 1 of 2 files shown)


aman-db (approved):
Looks good to me

gueniai (approved):

Issue #1207, opened 2024-05-13
Redefine views so that they are created from tables, not locations


gueniai:
Currently, views are created from the tables' path, which creates permissions issues with downstream consumers. What is the impact  regarding permissions?

PR #1241, opened 2024-06-19
Redefine views so that they are created from tables not locations

souravbaner-da:
Below Logic is implemented as part of View Creation:

- First Check if the ETL tables exists.

--  If table exists then create views from the ETL tables.
-- If table not exist then
---   Check if the path exist
------ If path does not exist then skip the view creation process.
------ If path does exist then create the table and do the normal view creation process.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ name := "overwatch"
 
 organization := "com.databricks.labs"
 
-version := "0.8.1.1"
+version := "0.8.2.0"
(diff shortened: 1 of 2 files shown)


aman-db (approved):
LGTM

neilbest-db (approved):

neilbest-db:
closes #1207 (right, @souravbaner-da ?)

PR #1255, opened 2024-07-02
1201 collect all event logs on first run

souravbaner-da:
For first run in overwatch cluster_events_bronze would now get all the records instead of some specific timestamp


sriram251-code (approved):

neilbest-db:
@souravbaner-da, I believe this is the one where you agreed to adjust the method signature and/or argument types for `landClusterEvents` in order to pass the code analysis check, correct?  You may request my review when that is ready.  Thanks.

src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala
neilbest-db, line 1: I added some whitespace here for legibility.

neilbest-db (approved):
LGTM.

neilbest-db:
This is ready to be merged but I am not yet able to do that.  I moved it to "Done" on the board but the PR icon there will remind us that it's open.

Issue #1258, opened 2024-07-09
Remove Duplicate Instance Types - AWS


mohanbaabu1996:
Remove Duplicate Instance Types: AWS

PR #1263, opened 2024-07-11
Update AWS_Instance_Details.csv

neilbest-db:
copied from https://github.com/mohanbaabu1996/overwatch/tree/duplicate_aws_instances

in lieu of #1259 attempting to pull from a forked repo.

PR #1282, opened 2024-08-28
fix duplicate AWS instances in 0.8.2.0

neilbest-db:
same as #1263 but needed for release testing.


neilbest-db:
(re-) closes #1258

gueniai (approved):

Issue #1284, opened 2024-09-05
Edge case for 1002 and 1005 failing Api call


ashlytee:
```
FAILED --> 
ERROR:
FAILED: 1005-Bronze_ClusterEventLogs Module: Failed

Unable to receive all the clusters/events api responses; Api response received
9285;
Api response not received 43775
java.lang.Exception: Unable to receive all the clusters/events api responses; Api response received
9285;
Api response not received 43775
```

We found that in one of the workspaces the cluster is not present hence no cluster activity. This is an edge case ( exception) that needs to be handled

PR #1285, opened 2024-09-09
Gracefully handle API call failure in cluster-events

souravbaner-da:
Gracefully handle API call failure in cluster-events

Solve #1284


neilbest-db (commented):
Would it be better from a separation-of-concerns perspective to simply allow this module to have the same `EMPTY` status message as any other module but indicate that there were API errors in the status of the upstream module?  if we are already doing that then it should be adequate unless we want to consider preventing downstream modules from proceeding in that scenario . . .  What's our established pattern in that respect?

An alternative to consider would be to automatically check for relevant records in `${erroredBronzeEventsTarget.tableFullName}` at runtime rather than expecting the user to notice this advice in the `pipReport`.

src/main/scala/com/databricks/labs/overwatch/pipeline/BronzeTransforms.scala
neilbest-db, line 1: Please see [this comment](https://github.com/databrickslabs/overwatch/pull/1285#pullrequestreview-2309443403) on the PR.

neilbest-db (commented):
Did you say that you are still working on more changes?  Maybe you have some commits to push?  LMK how I can help.

